PR #3, opened 2017-01-10
Fixed import path

soider:
There was obsolete import path via gopkg in readme, with wrong library name and version suffix.


davidbyttow:
Thank you @soider!

PR #19, opened 2017-11-04
Add support for compression and lossless options

mikestead:
Just a small addition to add the option of [Lossless](https://github.com/jcupitt/libvips/blob/master/libvips/foreign/webpsave.c#L367) compression for webp images.

While I was in there I also surfaced the `Compression(int)` option you'd added but which wasn't available through `Transform`.

I've had a test of these in my project and they both seem to be functioning ok, but I was having some issues getting your tests running locally so if you could give it the once over.

Let me know if you'd like any other tests.

---
This change is [Reviewable](https://reviewable.io/reviews/davidbyttow/govips/19)

### DIFF
--- a/bridge.c
+++ b/bridge.c
@@ -65,10 +65,11 @@ int save_png_buffer(VipsImage *in, void **buf, size_t *len, int strip, int compr
 	);
 }
 
-int save_webp_buffer(VipsImage *in, void **buf, size_t *len, int strip, int quality) {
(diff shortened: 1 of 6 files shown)


davidbyttow:
Feel free to merge.

mikestead:
@davidbyttow Afraid you'll need to, don't have write privileges :)

PR #25, opened 2018-01-25
Add rotation operation

Nafanya:
This change is [Reviewable](https://reviewable.io/reviews/davidbyttow/govips/25)


bridge.h
```diff
@@ -36,6 +36,10 @@ int remove_icc_profile(VipsImage *in);
 
 // Operations
 int resize_image(VipsImage *in, VipsImage **out, double scale, double vscale, int kernel);
+int rot_image(VipsImage *in, VipsImage **out, VipsAngle angle);
+int rot_image90(VipsImage *in, VipsImage **out);
```
davidbyttow: do we need these other methods versus just the single `rot_image` method where the angle is passed in?

transform.go
```diff
@@ -188,6 +189,30 @@ func (t *Transform) GaussBlur(sigma float64) *Transform {
...
 }
 
+// Rotate rotates image by a multiple of 90 degrees
+func (t *Transform) Rotate(angle Angle) *Transform {
+	t.tx.Rotate = angle
+	return t
+}
+
+// Rotate90 rotates image by 90 degress clockwise
+func (t *Transform) Rotate90() *Transform {
```
davidbyttow: I don't think we need these separate methods, just the `Rotate` with the given `Angle`

Nafanya:
Sounds reasonable, removed rotate by 90/180/270 methods and left the generic one

PR #47, opened 2018-08-28
Only unref if an image is not used as the source

davidbyttow:
This change is [Reviewable](https://reviewable.io/reviews/davidbyttow/govips/47)


oleh-ozimok:
@davidbyttow this fix produce memory leaks, need another solution

PR #55, opened 2018-11-28
Error while calling CropOffsetX and CropOffsetY with a negative value 

c93614:
Hi,

I found that if I use negative offset while calling `CropOffsetX()` & `CropOffsetY()`, the `left` and `top` value may calculate wrong.


tonimelisma:
Hey @c93614. Unfortunately due to a large amount of functionality we've merged from a fork and other updates, the code no longer merges - we no longer have that functionality. I'm closing the PR.

PR #26, opened 2018-01-26
WIP: allow image load options

victorkt:
This is a work in progress to add the ability of passing additional parameters to libvips loader functions, as mentioned on issue #23 . It also adds the options for other image types as well. It should already work, still pending the tests which I'll add later.

Please let me know your thoughts and what could be improved or changed.

---
This change is [Reviewable](https://reviewable.io/reviews/davidbyttow/govips/26)

PR #56, opened 2018-12-06
allow specifying access mode on image load functions

toaster:
This allows specifying a different access mode on the three image loading functions `NewImageFromBuffer`, `NewImageFromFile` and `LoadImage`.

The sequential access mode allows the processing of huge images without writing temp data to the filesystem. On an AWS Lambda there are only 500MB temp space which limits the size of images processable using the random access mode.
For straight forward image resizing or cropping the sequential mode is sufficient.

I decided to extend the signatures of the loading methods (which affects existing code) instead of adding another three functions. Existing code can easily be fixed by adding `nil` for the load options.
Wrapping the access mode specification in a load options type allows for future extensions (like the modes in #26) without breaking the API again.

The usage of the mode is tested by exporting a test image twice. In random access mode this works while it fails in sequential access mode.


pkg/vips/image.go
```diff
@@ -24,33 +24,38 @@ type ImageRef struct {
...
 }
 
+// LoadOptions contains optional configuration for the image loading.
+type LoadOptions struct {
+	Access Access
+}
+
 // LoadImage loads an ImageRef from the given reader
-func LoadImage(r io.Reader) (*ImageRef, error) {
+func LoadImage(r io.Reader, opts *LoadOptions) (*ImageRef, error) {
```
davidbyttow: Hm, I think i'd prefer the functor pattern, then there won't be any upstream changes needed at callsites. Thoughts?

```suggestion
func LoadImage(r io.Reader, opts ...LoadOption) (*ImageRef, error) {
```

Then it'd look like:
```golang
LoadImage(r, WithAccessMode(SequentialAccess))
```

Or something like that.

toaster: Yes, you're right.
I will change it to functors.

Issue #67, opened 2019-03-05
thumbnail support specify both height and width


leslie-wang:
Thanks for making the great project.

We want to specify both width and height when generating thumbnail because of evenly distributed tile screen. Can we add it?

PR #70, opened 2019-03-06
support thumnail generation with height parameter

leslie-wang:
also add one example application to generate thumbnail

close https://github.com/davidbyttow/govips/issues/67


leslie-wang:
seems like I should use option argument to achieve same goal, then doesn't require change

PR #77, opened 2019-03-27
Added support for OpenSlide formats.

freedombird9:
libvips support OpenSlide formats by default. The current binding initiates vips images only from buffer.  OpenSlide images cannot be loaded this way. I added several new APIs specifically to load OpenSlide images and images of any other formats that libvips possibly supports.

### DIFF
--- a/pkg/vips/bridge.c
+++ b/pkg/vips/bridge.c
@@ -35,7 +35,7 @@ int init_image(void *buf, size_t len, int imageType, ImageLoadOptions *o, VipsIm
 
 int init_open_slide(const char *filename, VipsImage **out) {
     int code = 1;
-    code = vips_openslideload(filename, out, NULL);
+    code = vips_vipsload(filename, out, NULL);
     return code;
 }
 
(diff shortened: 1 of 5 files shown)


tonimelisma:
@freedombird9 thanks for the PR! Unfortunately govips maintenance was on a bit of a pause although we're now getting back up to speed. I fully realise it's been over a year since you raised the PR - is it still relevant and should we work to get it to merge to the latest or should I just close the PR?

freedombird9:
@tonimelisma Well I don't have a preference since it's been so long and I've finished working on my project with my forked govip.  I think the feature is still nice to have though.

tonimelisma:
Ok, thanks for the info. Considering there's no immediately demand I won't prioritise it for now.

jonasteuwen:
Hi, I would still think openslide is a relevant feature. Could we add this feature?

tonimelisma:
Hey @jonasteuwen sure, if you'll provide an up to date PR that passes all the tests I'll review and merge it

PR #79, opened 2019-04-04
Add func NewImageFromNativeImage

duedal:
Uses vips_image_new_from_memory to load an image already in memory

I imagine the most common use case for govips is to convert/transform etc some image file and then save it to a different image file, but it would be nice to be able to operate on an image that's already in memory. Thus we can leverage govips to encode an image.Image, which we might have rendered, generated or loaded in some other way.

This PR is somewhat related to #2 - solving for the case where the image in memory is an image.Gray(16) or image.RGBA(64), or could be converted to it.

### DIFF
--- a/pkg/vips/image.go
+++ b/pkg/vips/image.go
@@ -6,6 +6,7 @@ import "C"
 
 import (
 	"errors"
+	nativeimage "image"
 	"io"
 	"io/ioutil"
 	"runtime"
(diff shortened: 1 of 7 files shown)


jhford-scout24:
@duedal nice PR, I really like the idea here.  I was wondering if you had thought about an opposite operation for getting an image.Image out from a vips.Image.  @ukautz and I are interested in this feature and can implement it.

tonimelisma:
Hey @duedal and @ukautz thanks a million for the PR. I like the idea and would welcome the functionality. Unfortunately we've merged a great amount of functionality from a fork which means the PR no longer merges. I'm trying to figure out if the ask is still relevant considering the issue hasn't been touched in over a year. Should I just close the PR or is there interest if I backport this to the latest?

jhford-scout24:
I don't think we're still looking to use this functionality, but it's a very interesting feature for the library.  We did use this PR in our forked repository for a while, and it worked as well as we'd hope it to.  I think it would be worth a backport, but I'm not sure if either myself or @ukautz are able to put in time on that.

tonimelisma:
Ok, thanks - probably if there's no immediate demand we won't prioritise this functionality to the top of the backlog.

PR #89, opened 2019-09-24
Suppress debug-level logs when required

### LINKED ISSUE #86: Suppress logging
Is there a way to suppress debug logging from the library, or control it somehow? It's quite annoying to see all the logs printed.

ItalyPaleAle:
Fixes #86

### DIFF
--- a/pkg/vips/govips.go
+++ b/pkg/vips/govips.go
@@ -29,6 +29,9 @@ const VipsMajorVersion = int(C.VIPS_MAJOR_VERSION)
 // VipsMinorVersion if the minor version of libvips
 const VipsMinorVersion = int(C.VIPS_MINOR_VERSION)
 
+// ShowDebugLog controls whether debug-level log is printed
+var ShowDebugLog = true
+
 var (
 	running           = false
 	initLock          sync.Mutex
--- a/pkg/vips/type.go
+++ b/pkg/vips/type.go
@@ -5,7 +5,6 @@ package vips
 import "C"
 import (
 	"io"
-	"log"
 	"strings"
 	"sync"
 )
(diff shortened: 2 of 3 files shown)


denisvmedia:
You can actually control the logs without modifying the library. Just change the log output:

```go
log.SetOutput(ioutil.Discard)
```

ItalyPaleAle:
My PR is a bit different because it doesn’t suppress *all* logs, but just debug ones. I still want to see errors

ItalyPaleAle:
@denisvmedia Following up on this... Any chance you could merge this?

Your workaround suppresses all logs, for the entire application. It operates in the global scope. I'd love it to just suppress the logs for govips and nothing else, and especially just have control over debug logs only.

As you can see between here and the other issue, there are upvotes by other too

denisvmedia:
Sorry to disapoint you, but no chances. I'm not the owner of the repo ;)

ItalyPaleAle:
That was.... awkward 😄  Sorry!!

@davidbyttow You can probably merge this!

pkg/vips/govips.go
```diff
@@ -29,6 +29,9 @@ const VipsMajorVersion = int(C.VIPS_MAJOR_VERSION)
 // VipsMinorVersion if the minor version of libvips
 const VipsMinorVersion = int(C.VIPS_MINOR_VERSION)
 
+// ShowDebugLog controls whether debug-level log is printed
+var ShowDebugLog = true
```
davidbyttow, line 33: This should probably default to false

davidbyttow:
But will merge as it doesn't change behavior.

PR #103, opened 2020-09-11
Add method for get EXIF orientation

bozaro:
Add method for get EXIF image orientation.
In my case it used for workaround https://github.com/davidbyttow/govips/issues/98

### DIFF
--- a/pkg/vips/bridge.c
+++ b/pkg/vips/bridge.c
@@ -37,6 +37,30 @@ VipsAngle autorot_get_angle(VipsImage *img) {
 	return vips_autorot_get_angle(img);
 }
 
+int exif_orientation(VipsImage *img) {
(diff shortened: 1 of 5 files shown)


davidbyttow (approved):
Tests would be nice too!

tonimelisma:
Hey @bozaro! Thanks so much for the PR. Unfortunately with the latest libvips version this information is impossible to get. Due to challenges in EXIF orientation, all EXIF files are autorated upon import.

Thanks for the PR, but we have migrated plenty of functionality from upstream which included, among other things, the feature to get metadata for any image. This metadata should have the orientation.

If I've misunderstood something, please let me know. Hope the newest govips version works for you!

bozaro:
Don't worry and thanks a lot for new API.
I'm already migrated to govips/v2 :+1:

PR #104, opened 2020-09-11
Add `ImageRef.Clone()` method

bozaro:
Allow multiple differ transformations for already loaded images.


davidbyttow (approved):
A test would be nice :)

PR #111, opened 2020-10-28
Add support for Magick loader and most loading parameters.

ibrt:
I was trying to process some animated GIFs - which are supported by libvips - but realized that doing so requires some changes in this library: 1) passing parameters to loaders, 2) extracting some additional metadata from images (e.g. the number of pages), and 3) supporting the magick writer.

I started taking a stab at (1), the only parameter that's missing is the "background" for PDF (I may add it - didn't want to figure out how it works just yet). While I was at it I also added the ability to use magick as loader.

I'll add tests of course, but I wanted to see if this PR may be accepted and agree on design first.

### DIFF
--- a/vips/foreign.c
+++ b/vips/foreign.c
@@ -1,32 +1,71 @@
 #include "lang.h"
 #include "foreign.h"
 
-int load_image_buffer(void *buf, size_t len, int imageType, VipsImage **out) {
-	int code = 1;
+int load_jpeg_buffer(void *buf, size_t len, VipsImage **out, int shrink, int fail, int autorotate) {
+	return vips_jpegload_buffer(buf, len, out,
+		"shrink", shrink,
+		"fail", INT_TO_GBOOLEAN(fail),
+		"autorotate", INT_TO_GBOOLEAN(autorotate),
+		NULL);
+}
 
-	if (imageType == JPEG) {
-		code = vips_jpegload_buffer(buf, len, out, NULL);
-	} else if (imageType == PNG) {
-		code = vips_pngload_buffer(buf, len, out, NULL);
-	} else if (imageType == WEBP) {
-		code = vips_webpload_buffer(buf, len, out, NULL);
-	} else if (imageType == TIFF) {
-		code = vips_tiffload_buffer(buf, len, out, NULL);
-	} else if (imageType == GIF) {
-		code = vips_gifload_buffer(buf, len, out, NULL);
-	} else if (imageType == PDF) {
-		code = vips_pdfload_buffer(buf, len, out, NULL);
-	} else if (imageType == SVG) {
-		code = vips_svgload_buffer(buf, len, out, NULL);
-	} else if (imageType == HEIF) {
-		// added autorotate on load as currently it addresses orientation issues
-		// https://github.com/libvips/libvips/pull/1680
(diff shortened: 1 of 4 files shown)


tonimelisma:
@ibrt looks great - thanks so much for the effort!

On the design itself, is there a reason we don't have the options struct as just directly editable from the host application, it looks like you have setter methods for each option? Perhaps we could simplify this, to be in line with how the export options also work currently?

@davidbyttow Any thoughts?

tonimelisma:
@ibrt also can you rebase to latest?

ibrt:
Couple reasons... One is that this way is backwards compatible while adding a struct parameter isn't. The other one is that the zero value of the struct wouldn't be initialized with the correct default values unless the client calls a NewDefaultImportOptions function (or similar), which I think is not ideal.

That said if you prefer the other design can change it.

tonimelisma:
Hey @ibrt thanks for the context.

On the backwards compatibility: I meant that we should make the ImportOptions.params struct public, perhaps moving all the importParams into public fields directly on the ImportOptions struct. Perhaps I'm being a bit slow here but I don't understand how this would affect backwards compatibility..?

Regarding the default values, I guess we have the exact same issue with the export params and this has been solved with Export(). There we use a pointer to the struct. I feel like generally it would be most logical if the import and export options functioned the same (perhaps even calling ImportOptions -> ImportParams). Would that work for the import parameters?

The HEIF autorotate thing is a bit of a tricky edge case but I think I have documented it in several places in the comments / godoc so if someone wants to shoot themselves in the foot by disabling HEIF autorotation they are free to do so 😆

ibrt:
The difference is that the `...ImportOption` arguments are optional, so the method signatures for `NewImageFromReader` etc. don't change in practice as they would if we added an `*ImportOptions` argument. Theoretically according to the Go module versioning spec this would require shipping a new major version. A middle-ground alternative is to add `NewImageFromReaderWithOptions` variants.

The issue with the export params design is that a) the zero value of the struct does not by default hold default values, and b) when we receive it we can't tell if any of the fields was explicitly set...

That said I see where you are going with this, I can change it to work in the same way as export params do.

tonimelisma:
Hey @ibrt that's perfect, I think we are on the same page. Govips was just a bit lacking in maintenance for a year or so and we just came out with 2.0.0, based on activity a lot of people are now trying it out so I'd definitely not want to break the API at this stage, nor go directly from 2.0.1 to 3.0.0.

So I think actually your middle ground seems like the perfect option to me.

tonimelisma:
Hey @ibrt in my other projects I've felt a great need for the fail parameter on loading so this PR would be greatly appreciated.

As discussed, the export parameters don't have any proper defaults either unless the user sets them, so I wouldn't be too worried at this stage. We can add a function to fill in the defaults (based on ImageRef) later, simultaneously for both import as well as export parameters.

Can I inquire about the progress?

Would be happy to include this in 2.1.0 along with the new logging functionality 😄

tonimelisma:
Hey @ibrt can I inquire about the progress on this PR?

tonimelisma:
We've now merged David Byttow's great work on loading parameters, it should probably suffice for this functionality. We still need to update various functions to support animated.

PR #114, opened 2020-11-02
Add GC support for `vips.ImageRef`

bozaro:
This code was already there.
At the moment, you need to explicitly execute `Close()`, otherwise there will be a memory leak.


tonimelisma:
Thanks @bozaro!

PR #134, opened 2020-12-10
Avoid allocating a new go byte slice buffer and reuse the C buffer on export

vansante:
This saves some quite large allocations when working with larger images.

I added an extra benchmark that shows this:

## Before:
```
BenchmarkExportImage-8   	      32	  35683545 ns/op	  377113 B/op	       6 allocs/op
```

## After:
```
BenchmarkExportImage-8   	      31	  35990512 ns/op	     280 B/op	       5 allocs/op
```

### DIFF
--- a/vips/foreign.go
+++ b/vips/foreign.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"image/png"
 	"math"
+	"reflect"
 	"runtime"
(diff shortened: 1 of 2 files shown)


tonimelisma:
Thanks @vansante! You're a big help!

PR #135, opened 2020-12-10
[WiP] Add VipsSourceCustom using go callbacks and an io.Reader

vansante:
Still needs some work.

### DIFF
--- a/vips/foreign.c
+++ b/vips/foreign.c
@@ -29,6 +29,11 @@ int load_image_buffer(void *buf, size_t len, int imageType, VipsImage **out) {
 	return code;
 }
 
+VipsImage * load_image_source(VipsSourceCustom *source) {
+	// https://libvips.github.io/libvips/API/current/VipsImage.html#vips-image-new-from-source
+	return vips_image_new_from_source( (VipsSource*) source, "", NULL);
+}
+
 // https://libvips.github.io/libvips/API/current/VipsForeignSave.html#vips-jpegsave-buffer
 int save_jpeg_buffer(VipsImage *in, void **buf, size_t *len, int strip, int quality, int interlace) {
     return vips_jpegsave_buffer(in, buf, len,
--- a/vips/foreign.go
+++ b/vips/foreign.go
@@ -179,6 +179,14 @@ func isBMP(buf []byte) bool {
 	return bytes.HasPrefix(buf, bmpHeader)
 }
 
+func vipsLoadSource(src *Source) (*C.VipsImage, error) {
+	img := C.load_image_source(src.src)
+	if img == nil {
+		return nil, fmt.Errorf("error loading image from source")
+	}
+	return img, nil
+}
+
 func vipsLoadFromBuffer(buf []byte) (*C.VipsImage, ImageType, error) {
 	src := buf
 	// Reference src here so it's not garbage collected during image initialization.
--- a/vips/foreign.h
+++ b/vips/foreign.h
@@ -20,6 +20,7 @@ enum types {
 };
 
 int load_image_buffer(void *buf, size_t len, int imageType, VipsImage **out);
+VipsImage * load_image_source(VipsSourceCustom *source);
 
 int save_jpeg_buffer(VipsImage* image, void **buf, size_t *len, int strip, int quality, int interlace);
 int save_png_buffer(VipsImage *in, void **buf, size_t *len, int strip, int compression, int interlace);
--- a/vips/image.go
+++ b/vips/image.go
@@ -100,12 +100,14 @@ func NewDefaultWEBPExportParams() *ExportParams {
 
 // NewImageFromReader loads an ImageRef from the given reader
 func NewImageFromReader(r io.Reader) (*ImageRef, error) {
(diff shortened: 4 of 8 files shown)


tonimelisma:
Thanks for this @vansante. Is it possible you explain a little bit about the objectives and benefits of the PR and the high-level architectural description how you are looking to implement this?

vansante:
@tonimelisma Yes, the idea is to make the reading of image truly streaming by using the new [libvips streaming api](https://libvips.github.io/libvips/2019/12/11/What's-new-in-8.9.html). So then there will be a technical difference between opening an image using `NewImageFromReader` which utilizes an `io.Reader` to load the image data directly into vips and using `NewImageFromBuffer` which loads a binary blob into memory and then passes that to vips (currently they do the same thing, except `NewImageFromReader` uses `ioutil.ReadAll` to get the data). 

The idea is to also introduce the streaming output `VipsTargetCustom` in a subsequent MR so you can use vips fully via its streaming api.

Let me know if this is a desired change before I proceed :)

tonimelisma:
Sounds good to me - this would be an exciting upgrade!

I'm wondering what all the implications to the core functionality will be. Will all core processing methods need to be upgraded? Will the core processing methods automatically detect whether they're working with a buffer or stream?

@davidbyttow @Elad-Laufer  what are your thoughts? Do you see any issues or do you have guidance how to implement this?

Thank you @vansante for being an active contributor!

zzjin:
Great work!
I've made some benchmarks witch shows maybe there was not enough difference:

bench test code:
```go
var testFile = resources + "jpg-24bit.jpg"
var testFunc = func(img *ImageRef) {
	img.Thumbnail(100, 100, InterestingCentre)
}

func BenchmarkLoadFromFile(b *testing.B) {
	benchHelper(b, testFile, NewImageFromFile, testFunc)
}
func BenchmarkLoadFromSource(b *testing.B) {
	benchHelper(b, testFile, func(file string) (*ImageRef, error) {
		// Test with using an image source
		imageFile, _ := os.Open(file)
		return NewImageFromReader(imageFile)
	}, testFunc)
}
```

witch output is:
```
goos: linux
goarch: amd64
pkg: github.com/davidbyttow/govips/v2/vips
BenchmarkLoadFromFile-8     	     339	   3502284 ns/op	    3956 B/op	      29 allocs/op
BenchmarkLoadFromSource-8   	     276	   3798112 ns/op	    2685 B/op	      47 allocs/op
```
test with thumbnail func, which is most common usage.

PS: libvips mem shows:
```
file: Mem: 30320640, MemHigh: 30374736, Files: 0, Allocs: 5040
```

```
source: Mem: 33569280, MemHigh: 33623376, Files: 0, Allocs: 5580
```

sethify:
This is exciting. Is this PR still active?

tonimelisma:
Hey @vansante! This PR still looks awesome. Do you see any technical or architectural issues blocking progress? Do you feel like you could allocate time for this in the coming weeks?

vansante:
I was not really sure if this was still something to be desired after zzjin mentioned that the benchmarks are not showing improvements so I had not looked at it anymore since then. I should probably do some benchmarking of my own (possibly with the writers streaming as well). I think I can take a look at it again in the next 2 weeks.

tonimelisma:
Hmm. I think the original issue where the benchmark was raised related to throughput speed, which the benchmark is actually not measuring. Govips is used in several backend services where streaming support might provide big end-to-end throughput benefits, I'm all for it.

vansante:
I had a look at the current MR and I identified the following issues:

- All images loaded through the streaming `VipsSourceCustom` loader seem to lose metadata (EXIF and stuff), and thus most of the Golden image tests fail.
- I made the `free()` function on the `Source` struct, but its never called. I just tried creating a third callback for `close` but that doesnt work. I am not quite sure when it is safe to garbage collect the C `GoSourceArguments` struct and remove the `Source` from the map in Go.
- I would rather have a direct pointer in C-land in the `GoSourceArguments` struct rather than keeping and manager a global map with associated lock in Go. This would make the setup much more elegant.

If anyone has any good ideas for these issues, please let me know! :)

tonimelisma:
Apologies, but I'm doing some cleanup and closing this PR due to lack of activity. I still think the streaming API would be a great improvement but unfortunately don't have great ideas how to architect this change. Thanks for the contribution in any case!

negator:
@vansante Have you considered using the [Pipe function](https://pkg.go.dev/os#Pipe), which returns a pair of files for reading/writing? You can access the underlying file descriptors to create VipsSource and VipsTarget objects (using [vips-source-new-from-descriptor](https://www.libvips.org/API/current/VipsTargetCustom.html#vips-source-new-from-descriptor) and [vips-target-new-to-descriptor](https://www.libvips.org/API/current/VipsTargetCustom.html#vips-target-new-to-descriptor)), and forgo the need for `GoSourceArguments` and the global map entirely.

vansante:
Hi @negator,

To be honest I haven't looked (much) at the (go)vips code since my last comment, and I am not sure whether I will have time to look at this again soon. Your suggestion sounds very promising however, so feel free to take over from me!

PR #137, opened 2020-12-13
Revert "Avoid allocating a new go byte slice buffer and instead reuse the C buffer"

vansante:
This reverts commit bf2aa07cac26c4995a3bdfcf6e5c61adefedc98a.

### DIFF
--- a/vips/foreign.go
+++ b/vips/foreign.go
@@ -9,7 +9,6 @@ import (
 	"fmt"
 	"image/png"
 	"math"
-	"reflect"
 	"runtime"
 	"unsafe"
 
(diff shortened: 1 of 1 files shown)


tonimelisma:
Hey @vansante - are you sure your PR is the reason for the behavior? If you check the attached log you'll see that Go garbage collection is indeed calling imageRef.close(). 

```
% grep -i created.imageref vips_log.txt | wc -l
      42
% grep -i closing.image vips_log.txt | wc -l
      11
```

So I think the problem is actually in the close() function, not in your patch? What do you think?

tonimelisma:
Thanks @vansante you're right, I got rid of the memory leaks by reverting the commit. However, the fact is govips is still using huge amounts of memory which is a different issue to solve.

PR #139, opened 2020-12-14
implement divide

axkirillov:


axkirillov:
My divide test is failing but its not on my machine, while other tests, like multiply fail on my machine. If I delete golden image for multiply and generate a new one, the test passes on my machine. So this is probably an issue with my setup, and the golden image I produce on my mac doest pass in the build.

axkirillov:
@tonimelisma any ideas on why could that be?

tonimelisma:
@axkirillov that's likely due to a weird version of libvips or something similar. Please submit the PR without including the golden picture in the resources folder.

axkirillov:
@axkirillov i just did brew reinstall vips and I still produce wrong images, here you can see that color is different
image

axkirillov:
@tonimelisma I deleted the image and now it passes

tonimelisma:
Generally looks great. The relative test coverage on the total shows it's decreasing a little bit, is it possible that you fix that?

PR #140, opened 2020-12-14
implement mapim

axkirillov:


tonimelisma:
Thank you for the three PRs @axkirillov, these are greatly appreciated!

Is it possible you write golden unit tests for each of the three PRs in the image_golden_test.go file? You can get a good example e.g. from TestImage_ResizeWithVScale(). Then if you run the tests twice it will also produce a golden comparison image in the resources subdirectory which you should attach to the PR.

Thanks!

axkirillov:
> 
> 
> Thank you for the three PRs @axkirillov, these are greatly appreciated!
> 
> Is it possible you write golden unit tests for each of the three PRs in the image_golden_test.go file? You can get a good example e.g. from TestImage_ResizeWithVScale(). Then if you run the tests twice it will also produce a golden comparison image in the resources subdirectory which you should attach to the PR.
> 
> Thanks!

Ok, will do!

tonimelisma:
Apologies for the strictness with the unit tests, but we've just achieved 80% test coverage and we need to maintain it or increase it a little bit with all changes in order to be accepted as a mature library in different directories. You can see the test coverage in the coveralls comment above. Consider also writing failing tests so you cover 100% of the code you're contributing, or perhaps adding a test somewhere else over uncovered code to compensate if you can't test 100% of your own code.

axkirillov:
> 
> 
> Apologies for the strictness with the unit tests, but we've just achieved 80% test coverage and we need to maintain it or increase it a little bit with all changes in order to be accepted as a mature library in different directories. You can see the test coverage in the coveralls comment above. Consider also writing failing tests so you cover 100% of the code you're contributing, or perhaps adding a test somewhere else over uncovered code to compensate if you can't test 100% of your own code.

No problem! I didn't write unit tests at first because these functions are not public facing and are meant to be used internally. Plus they just provide an interface to vips without any logic of their own. But its a good idea, I can just add tests for something that is not yet covered to bump up the number =)

axkirillov:
@tonimelisma I updated all prs so that all new code is covered by tests

PR #145, opened 2020-12-20
Fix panic on shutdown

davidbyttow:
Upon termination, if libvips continues work in another thread (such as for reporting leaks) and logs, then a panic will occur if the log handler is a callback into Go. This is not an ideal fix because all processes/threads should terminate appropriately, but it's a stopgap to prevent the failure.

There may be a better way to handle this (via a signal callback or something else) to catch it just as Go is exiting and unset the handler.

Users are suggested to call Shutdown() upon exit to avoid this error.

### DIFF
--- a/vips/govips.go
+++ b/vips/govips.go
@@ -86,6 +86,7 @@ func Startup(config *Config) {
 	// Initialize govips logging handler and verbosity filter to historical default
 	if !currentLoggingOverridden {
 		LoggingSettings(nil, LogLevelInfo)
+		currentLoggingOverridden = false
 	}
 
 	// Override default glib logging handler to intercept logging messages
--- a/vips/govips.go
+++ b/vips/govips.go
@@ -35,6 +35,7 @@ const (
 
 var (
 	running             = false
+	hasShutdown         = false
 	initLock            sync.Mutex
 	statCollectorDone   chan struct{}
 	once                sync.Once
@@ -56,9 +57,14 @@ type Config struct {
 // Startup sets up the libvips support and ensures the versions are correct. Pass in nil for
 // default configuration.
 func Startup(config *Config) {
+	if hasShutdown {
+		panic("govips cannot be stopped and restarted")
+	}
+
 	initLock.Lock()
-	runtime.LockOSThread()
 	defer initLock.Unlock()
+
+	runtime.LockOSThread()
 	defer runtime.UnlockOSThread()
 
 	if running {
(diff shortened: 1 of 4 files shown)


vips/govips.go
```diff
@@ -86,6 +86,7 @@ func Startup(config *Config) {
 	// Initialize govips logging handler and verbosity filter to historical default
 	if !currentLoggingOverridden {
 		LoggingSettings(nil, LogLevelInfo)
+		currentLoggingOverridden = false
```
davidbyttow, line 89: Isn't this already false?

davidbyttow: oh I see, nevermind. Ugh, this is a little messy, but OK

tonimelisma: Sorry 😄 You get what you pay for 😉

tonimelisma: The null logging handler is the same technique Libvips uses in C to shut down logging. Not sure if your method works just as well but wanted to be safe. However it seems like the macOS tests are still failing so I guess it wasn't this?

davidbyttow: Hmm, I was able to repro locally (by making some changes) and then this fixed it. Let me see

davidbyttow: I think it's the non-deterministic concurrency of the tests. Also the tests never call Shutdown explicitly so, it's not to be expected that the tests will pass in every environment. By default parallelism is set to MAXPROCS, which will vary obviously

davidbyttow: Unfortunately, we can't call shutdown in tests because libvips does not support shutdown/startup in the same process. Cleanest way I could imagine is to call to manually call our tests as subtests from a single test and shutdown there.

tonimelisma: Oh man. And Go doesn't have atexit either. I guess we could set MAXPROCS to 1, but is test order execution guaranteed any way? And the macOS CI builds take like 15 minutes already. I'd really like to at least get the CI builds working as they're quite useful...

davidbyttow: Well, I think my suggestion may be what we're left with.

Or there's something useful in here: https://golang.org/pkg/os/signal/

tonimelisma: I don't think a program exit can raise a signal, it's only the other way around. I consider this issue purely a CI issue, as you said users should call Shutdown(). Also, this is just interim, as there's a bug also open for the Go runtime for macOS and I hope they fix that someday.

So the simple option is to just disable the logging handler for the unit tests when they're running in Github CI. I'll just write a patch for that.

davidbyttow: That seems reasonable.

tonimelisma: Done. Thanks for the help David.

Issue #21, opened 2017-12-12
Resized image file size 20% to 40% larger than "sharp" of Node.js


sann1918:
govips produces 20% - 40% larger image file sizes than sharp (node equivalent) does for the same image dimensions and the same Lanczos3 kernel. 

Please check if govips is producing the optimal file size since both are using the same libvips backend.

Thanks.

davidbyttow:
Interesting, thanks for that. Can you give an example image or is it consistent across all images?

sann1918:
It's resizing ordinary iPhone photo images. Photos are 4K, 14 ~ 17 MB each. After running 20 photos, around 10 photos have "the larger image size issue". If you need one image, please let me know how to get it to you. It's 17 MB.

sann1918:
David, I found your gmail on Medium. Just sent you the original image and the resized ones. One by govips and the other by sharp. 

the original 15 MB (4032 x 3024).
govips produces a 1.3 MB file (806 x 605)
sharp produces a 877KB file (806 x 605)

Please check gmail and let me know if you need more info.

davidbyttow:
Looking at this now

davidbyttow:
Hey @sann1918, can you show me your sharp code? I'm resizing like this but getting 1.3MB, which is the same as VIPs

```
sharp = require('sharp');

sharp('onions.png')
  .resize(806, 605)
  .toFile('onions-sharp.png');
```

sann1918:
I double checked. The results is 877 KB. 

This is the code just ran on node v8.9.3 on the latest macOS 

```javascript
const fs = require('fs');
const sharp = require('sharp');

fs.readdir('./images/', (err, files) => {
    files.forEach(file => {
        sharp('./images/' + file)
            .resize(806, 605)
            .toFile('./resized/' + file);
    });
});
```

devmarco:
Any update on that?

I'm resizing an image to 900/900, converting to `TIFF` with quality=100

```
vips.NewTransform().
    LoadBuffer(imageBody).
    Resize(900, 900).
    Format(vips.ImageTypeTIFF).
    Quality(100).
    StripMetadata().
    OutputBytes().
    Apply()
```

The generated image has the size of `2.5mb`, and the same file generated using Sharp (Nodejs version) has `852kb`, its a huge difference

tonimelisma:
[...]

PR #146, opened 2020-12-29
Fixes TIFF compression for #21

davidbyttow:


tonimelisma (commented):
Not sure how much difference it makes but the Go examples have a specific format, they're built as part of testing etc. [Testable Examples](https://blog.golang.org/examples) shows the package to be ```package_test```

davidbyttow:
> Not sure how much difference it makes but the Go examples have a specific format, they're built as part of testing etc. [Testable Examples](https://blog.golang.org/examples) shows the package to be `package_test`

Yeah, I re-added it back.

Also, how did you generate the golden files for the other platforms?

davidbyttow:
Also, do these qualify as examples? These just seem like example programs with main functions (not Example tests)

tonimelisma:
I've got a bunch of docker images with different Ubuntu versions. I'll build the golden images.

I guess we can keep them in examples if we want to, although we could put them in tests as benchmarks as well which have their own format (not the JS one though).

One question: should we also create a NewDefaultTIFFExportParams() function to create default values for these new parameters or is NewDefaultExportParams() sufficient?

We should also probably work to finish the PR about the import parameters.

Otherwise looks good to me, I can merge after adding the golden.

davidbyttow:
I think NewDefaultExportParams is sufficient for now. However, I also think we will want to move toward format-specific export types. I haven't given too much thought yet on how to do that, but I like how Sharp has explicit output functions that allow you to pass in all of the format-specific options. We should offer the same flexibility.

PR #149, opened 2021-01-02
Vips similarity, bandjoin_const & smartcrop

boosh:
Adds [vips_similarity](http://libvips.github.io/libvips/API/current/libvips-resample.html#vips-similarity). It supports images both with and without alpha channels. Note: I haven't added the ability to pass the `interpolate` parameter. That can be a future enhancement.

I've also added [bandjoin_const](http://libvips.github.io/libvips/API/current/libvips-conversion.html#vips-bandjoin-const) to allow adding an alpha channel (then I saw there was already an `AddAlpha`), but I've done it now so might as well leave it in.

Also added a function to export vips images as golang image.Image objects.

Like I say, I'm not a C programmer, so please check for memory leaks.

Thanks


tonimelisma:
Hey @boosh solid work, thank you so much for the contribution! I'll check for the memory leaks.

One thing - can you create unit tests? Positive tests in the golden tests file and any major failing tests in the regular one?

tonimelisma:
Looks solid through and through. Excellent work!

tonimelisma:
Actually one comment regarding the ToImage() function - I love the idea, but I'm not sure if the current code works. I think the Decode() function needs to initialise each Go library for each image type it should decode. So we both need to import lots of different image libraries, and generally check the image type before ToImage() in case the image type isn't supported.

boosh:
It looks like importing a package for each image type that `ToImage` can work with isn't required. I've added tests an everything LGTM :-)

tonimelisma:
Thanks a million @boosh. Hope you stick around and contribute more!

Issue #147, opened 2020-12-29
Unexpected behaviour on loading truncated images


bozaro:
Now call `NewImageFromBuffer` for truncated `jpeg` and `png` load successfully and produce some warning. It was unexpected for me.

Warnings are useless in this case: in a multithreaded application, they cannot be associated with a specific function call. I expected an error like `io.ErrUnexpectedEOF`.

This behavior can be controlled through the flag `"fail"` (with the exception of https://github.com/libvips/libvips/issues/1946).

For example:
```diff --git a/vips/foreign.c b/vips/foreign.c
index e3f3f17..169e371 100644
--- a/vips/foreign.c
+++ b/vips/foreign.c
@@ -5,9 +5,9 @@ int load_image_buffer(void *buf, size_t len, int imageType, VipsImage **out) {
        int code = 1;
 
        if (imageType == JPEG) {
-               code = vips_jpegload_buffer(buf, len, out, NULL);
+               code = vips_jpegload_buffer(buf, len, out, "fail", TRUE, NULL);
        } else if (imageType == PNG) {
-               code = vips_pngload_buffer(buf, len, out, NULL);
+               code = vips_pngload_buffer(buf, len, out, "fail", TRUE,  NULL);
        } else if (imageType == WEBP) {
                code = vips_webpload_buffer(buf, len, out, NULL);
        } else if (imageType == TIFF) {
```

I admit the current behavior might be helpful. And the value of the `"fail"` flag should be controlled by the user.

I see the following solutions to the problem:
 - Add `LoadingOptions` parameter like `func NewImageFromBuffer(buf []byte, opts *LoadingOptions) (*ImageRef, error) {` (but this is breaking change);
 - Add `LoadingOptions` parameter like `func NewImageFromBuffer(buf []byte, opts ...LoadingOptions) (*ImageRef, error) {` with `MergeLoadingOptions` method like https://github.com/mongodb/mongo-go-driver/blob/master/mongo/options/dboptions.go (but this is minor breaking change);
 - Add global variables for `fail` flag (but this is ugly solution, especially in a multi-threaded application).

I would like to know which path is preferable before submitting the PR with changes.
 
P. S. https://twitter.com/kernio/status/879944532808278016

tonimelisma:
Hey @bozaro thanks for this. There is already a more generic PR being worked on here: https://github.com/davidbyttow/govips/pull/111

[...]

PR #155, opened 2021-01-04
ImportParams proof of concept

bozaro:
Related to https://github.com/davidbyttow/govips/issues/147

I created `ImportParams` struct with fields:
 - `Fail`
 - `AutoRotate`
 - `Shrink`

All of them is optional and by default don't set on `...load_buffer` call.
By the same way can be added any other optional field.

### DIFF
--- a/vips/foreign.c
+++ b/vips/foreign.c
@@ -1,30 +1,92 @@
 #include "foreign.h"
-
+#include "govips.h"
 #include "lang.h"
 
-int load_image_buffer(void *buf, size_t len, int imageType, VipsImage **out) {
+int load_image_buffer_operation(VipsOperation *operation, VipsBlob *blob, VipsImage **out, ImportParams *params) {
+  int result;
+
+  // Set operation parameters
+  result = vips_object_set(VIPS_OBJECT(operation), "buffer", blob, "out", out, NULL);
+  if (result) {
+    return result;
+  }
+
+  if (params->fail_set) {
+    result = vips_object_set(VIPS_OBJECT(operation), "fail", params->fail, NULL);
+    if (result) {
+      govipsLoggingHandler("govips", G_LOG_LEVEL_INFO, "can't set optional parameter \"fail\"");
+    }
+  }
+
+  if (params->autorotate_set) {
+    result = vips_object_set(VIPS_OBJECT(operation), "autorotate", params->autorotate, NULL);
+    if (result) {
+      govipsLoggingHandler("govips", G_LOG_LEVEL_INFO, "can't set optional parameter \"autorotate\"");
+    }
+  }
+
+  if (params->shrink_set) {
+    result = vips_object_set(VIPS_OBJECT(operation), "shrink", params->shrink, NULL);
+    if (result) {
+      return result;
+    }
+  }
+
+  // Execute operation
+  result = vips_cache_operation_buildp(&operation);
+  if (result) {
+    return result;
+  }
+
+  // Get operation output
+  g_object_get(VIPS_OBJECT(operation), "out", out, NULL);
+  return 0;
+}
+
+int load_image_buffer_generic(const char *method, void *buf, size_t len, VipsImage **out, ImportParams *params) {
+  VipsOperation *operation;
+  VipsBlob *blob;
+  int result;
+
+  operation = vips_operation_new(method);
+  if (operation == NULL) {
+    return -1;
+  }
+
+  blob = vips_blob_new(NULL, buf, len);
+  result = load_image_buffer_operation(operation, blob, out, params);
+
+  vips_area_unref(VIPS_AREA(blob));
+  vips_object_unref_outputs(VIPS_OBJECT(operation));
+  g_object_unref(operation);
+  return result;
+}
+
+int load_image_buffer(void *buf, size_t len, int imageType, VipsImage **out, ImportParams *params) {
   int code = 1;
 
   if (imageType == JPEG) {
-    code = vips_jpegload_buffer(buf, len, out, NULL);
+    code = load_image_buffer_generic("jpegload_buffer", buf, len, out, params);
   } else if (imageType == PNG) {
-    code = vips_pngload_buffer(buf, len, out, NULL);
+    code = load_image_buffer_generic("pngload_buffer", buf, len, out, params);
   } else if (imageType == WEBP) {
-    code = vips_webpload_buffer(buf, len, out, NULL);
+    code = load_image_buffer_generic("webpload_buffer", buf, len, out, params);
   } else if (imageType == TIFF) {
-    code = vips_tiffload_buffer(buf, len, out, NULL);
+    code = load_image_buffer_generic("tiffload_buffer", buf, len, out, params);
   } else if (imageType == GIF) {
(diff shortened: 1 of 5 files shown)


tonimelisma:
Thanks a lot for these @bozaro! I think these look good to me. @davidbyttow had some thoughts on restructuring the export parameters. David, do you have any input into this import parameter PR?

The import parameters are greatly needed but I think it would be best if the import parameter behaviour matched the export parameters. Did you David already have an idea of deprecating the current export parameter behaviour?

Maybe it makes sense to align on this before bozaro puts more work into the PR.

tonimelisma:
Hey @bozaro I think it makes sense to pause a little bit to get David's guidance on how to move forward.

Now that I'm looking at the commits more closely, it looks like you're changing the API in a breaking fashion. Unfortunately we can't do this, we need to be backwards compatible in govips 2.x. Is it possible if you follow the discussion in https://github.com/davidbyttow/govips/pull/111 to look at how this might be implemented in a backwards compatible fashion?

That would mean:
- Leaving NewImageFromBuffer() and other current functions as they are right now (so we don't break compatibility)
- Adding NewImageFromBufferWithOptions() or something similar which would be the new function, that uses the new ImportOptions struct
- Treat the ImportOptions struct just like the ExportOptions struct. So instead of methods we would have a function to return imagetype-specific import options and generic import options, the user could change what they want and pass that to the new function

bozaro:
> - Leaving NewImageFromBuffer() and other current functions as they are right now (so we don't break compatibility)
> - Adding NewImageFromBufferWithOptions() or something similar which would be the new function, that uses the new ImportOptions struct

Agreed.

bozaro:
> Treat the ImportOptions struct just like the ExportOptions struct. So instead of methods we would have a function to return imagetype-specific import options and generic import options, the user could change what they want and pass that to the new function

I create `ImportParams` structure with this behaviour to avoid the following problems:

 - I use pointers to reserve `nil` as `default value`. This decision allow don't copy default from libvips code and avoid out-of-sync between govips and libvips defaults;
 - I use pointers to allow skip pass options to load function. This decision allow to remove information about which flags support the used formats from govips (it not so trivial information, for example: `fail` undocumentated for `png` but supported, documenated for `jpeg` but don't work);
 - Method `NewDefaultPNGImportParams` require knowledge about source format. In this case I can ignore some unsupported options safely (for example `fail` flag);
 - Pointer fields is not comfortable to use in struct initilization (I can't write `&true`) for this i added builder methods.

vips/foreign.h
```diff
@@ -19,7 +19,18 @@ enum types {
 	BMP
 };
 
-int load_image_buffer(void *buf, size_t len, int imageType, VipsImage **out);
+typedef struct {
+    gboolean fail_set;
```
davidbyttow, line 23: Why did you decide to have separate variables vs just using pointers with null checks?

vips/foreign.c
```diff
@@ -1,29 +1,92 @@
...
+int load_image_buffer_operation(VipsOperation *operation, VipsBlob *blob, VipsImage **out, ImportParams *params) {
+    int result;
+
+    // Set operation parameters
+    result = vips_object_set(VIPS_OBJECT(operation), "buffer", blob, "out", out, NULL);
+    if (result) {
+        return result;
+    }
+
+    if (params->fail_set) {
```
davidbyttow, line 14: This repeating pattern can be DRYed up in a helper function to handle logging if it fails.

But when does this sort of things fail?

vips/foreign.c
```diff
@@ -1,29 +1,92 @@
...
+#include "govips.h"
 
-int load_image_buffer(void *buf, size_t len, int imageType, VipsImage **out) {
+int load_image_buffer_operation(VipsOperation *operation, VipsBlob *blob, VipsImage **out, ImportParams *params) {
+    int result;
+
+    // Set operation parameters
+    result = vips_object_set(VIPS_OBJECT(operation), "buffer", blob, "out", out, NULL);
+    if (result) {
+        return result;
```
davidbyttow, line 11: why no log for this?

vips/foreign.c
```diff
@@ -1,29 +1,92 @@
...
+    }
+
+    if (params->fail_set) {
+        result = vips_object_set(VIPS_OBJECT(operation), "fail", params->fail, NULL);
+        if (result) {
+            govipsLoggingHandler("govips", G_LOG_LEVEL_INFO, "can't set optional parameter \"fail\"");
```
davidbyttow, line 17: Are you sure we want to carry on if this fails? In what cases does it fail? If I load an image with autorotate, and it fails to carry it out, this would just log and carry on without the user code being able to deal w/ it. Why not just fail altogethre?

vips/foreign.h
bozaro: I'm not sure if I can put together the correct construct with a pointer at the bridge of Go and C.

vips/foreign.c
bozaro: I can use `#define` for this block if you want to.

vips/foreign.c
bozaro: The corresponding message will be inside the error.

vips/foreign.c
bozaro: As I understand it, the fail flag is not supported only if partial image loading is supported. In this case, if it is not supported, then it can simply be ignored.
In such a situation, loading images will behave as if this flag is true.

davidbyttow:
Hey @bozaro, again thanks for your work on this. Check this PR out and see if it satisfies your requirements: https://github.com/davidbyttow/govips/pull/166

bozaro:
#166 contains a problem that I tried with all my might to avoid: discrepancy between `libvips`, `govips` and `libvips` documentation flags.
For example: `fail` flag supported at least for `vips_jpegload_buffer` (documented, but broken) and `vips_pngload_buffer` (undocumented, but works and lost in #166).

bozaro:
What decision will be about this PR:
 - Rebase, replace copy-paste by `#define` and merge?
 - Close and forget?
 - Something else?

vips/foreign.c
davidbyttow: Yeah, that may be better.

davidbyttow (commented):
Given my other PR (which is based on yours), I think it's worth just moving over to that and providing your feedback as a reviewer here: https://github.com/davidbyttow/govips/pull/166

Would love your feedback/suggestions @bozaro

vips/image.go
```diff
@@ -289,6 +289,7 @@ func finalizeImage(ref *ImageRef) {
 // Deprecated: Please remove all Close() functions from your code as superfluous.
 func (r *ImageRef) Close() {
 	govipsLog("govips", LogLevelInfo, "Close() is a deprecated function. You can remove it from your code as unnecessary.")
+	r.close()
```
davidbyttow, line 292: thanks!

davidbyttow:
I need to pick this back up, and will do so in the next day or so

davidbyttow:
hey @bozaro checkout https://github.com/davidbyttow/govips/pull/166

bozaro:
Thanks a lot!

PR #167, opened 2021-01-14
add ushort parameter to Identity

axkirillov:
this PR adds a a ushort parameter to the Identity call. When set to true the resulting LUT will be 16 bit wide

https://libvips.github.io/libvips/API/current/libvips-create.html#vips-identity


tonimelisma:
@axkirillov the C code doesn't compile

axkirillov:
> @axkirillov the C code doesn't compile

sorry, committed too fast without checking first xD

should be working just fine now

PR #170, opened 2021-01-18
Uses libvips operation API for saving

davidbyttow:


davidbyttow:
FYI @tonimelisma, I still need to thread through the optionals, but this mostly demonstrates how I'm thinking about invoking load/save

PR #173, opened 2021-02-10
Add mozjpeg equivalent parameters to save

sandro:
jcupitt explains the mozjpeg parameters in this comment:
https://github.com/libvips/libvips/issues/1422#issuecomment-530032329

Additionally, look for references to mozjpeg in the documentation for
vips_jpegsave():
https://libvips.github.io/libvips/API/current/VipsForeignSave.html#VipsForeignJpegSubsample

These changes improved jpeg compression for my test image, especially
with the addition of overshoot_deringing when compressing a screenshot
that includes text.


tonimelisma:
Hey @sandro thank you so much for the PR! The code looks great!

A couple of thoughts:
- You should note that we'll likely move into completely new export functions later but in the interim this is a good PR. See PR #166 for more info
- I'm all for adding the new parameters for JPEG exports, this is great work
- Is it possible you add or modify unit tests (especially golden tests) to cover any new parameters?

However, on the topic of NewMozJpegExportParams() I'm thinking this is more of a stylistic choice that should rather be covered in the documentation instead of creating a new default export template function.

Right now govips is just a minimal library of bindings on top of libvips, and we provide the parameters developers need to adjust exports, as well as a set of sane defaults if you're not bothered. Creating different templates for different export cases might be a possible future direction, but would need an intentional design that doesn't clutter the current API too much. So in essence if we had templates, maybe you would just have a constant parameter which you supply the ExportParams() functions such as Export(buf, govips.ExportTemplateScreenshot), Export(but, govips.ExportTemplateMozillaDefaults), etc.

How do you feel if instead of the NewMozJpegExportParams() functions you just created an example Go program in the examples/ subfolder, which manually set the corresponding parameters on top of NewJpegExportParams() and then add a notification to the GoDocs of the NewJpegExportParams() function that mentions the recommended parameters and the example program?

Does it make sense to you?

sandro:
@tonimelisma Thank you for your supportive words and thorough response!

I'll be happy to update the unit tests, thanks for pointing them out.

Yes, I agree that NewMozJpegExportParams() clutters up the API, I just wasn't sure of a better way to communicate the results of my digging through source code and issues to find how to replicate mozjpeg with vips. I didn't think my PR would get any attention, so I created this very loud function to help anyone who may have been searching PRs for references to mozjpeg.

I like the idea of having different export templates for different scenarios in the future! A few strong opinions could really help image compression newbies like myself. In the meantime, I'll add an example as you suggested.

Thanks

sandro:
@tonimelisma I force pushed my branch with the addition of a few golden tests for the new params and an example documenting the mozjpeg params.

I wasn't sure how best to test the new jpeg params, so I'm testing the buffer output size against the size of the file I get when running `vips jpegsave` locally. I added the vips command I used as a comment before each test case.

Feel free to rename, and move things around as necessary to make them best fit into the project.

Thank you,
Sandro

tonimelisma:
Great work. You put a lot of effort here. The additional parameters will be very useful! The automated tests are failing however, it seems the golden image tests don't match, either on Ubuntu 18.04 or 20.04. How did you produce the test images, did you use the latest libvips 8.10.2-1 from my PPA?

Thanks again for the great contribution!

sandro:
@tonimelisma Thank you for the assistance! No, I did not use your PPA, I didn't know what that was until a few minutes ago. I developed the changes locally on my mac using vips 8.10.5_2 installed via homebrew.

I'd be happy to modify the tests so they resemble the other golden tests. I performed the length check because I wasn't 100% sure that the test code I wrote actually worked. If I had not performed the length check and submitted my first version of my tests, a golden image would have been created though it would have been wrong because I had not set the `FormatType` to jpeg. The code ran without error, but the export parameters were not used and I would have committed a false positive to the test suite. Now that I have some certainty that the tests work (after the length check which compares against running the `vips jpegsave` command locally), I feel better about removing the length check. I still believe the golden tests could benefit from testing against the output of `vips` commands directly, but that's another discussion.

The contribution guide isn't clear on using or not using the PPA or ubuntu to develop changes. Would you recommend me creating a Docker image to generate the correct golden images?

Thank you

tonimelisma:
Oh, right, I thought you were developing on Ubuntu but you're using macOS. Yeah the golden reference images need to be created either natively in the environment or within docker as the compiled libvips libraries differ so much between different environments on both Ubuntu as well as macOS.

Listen, just create the golden images for the environment you have (either macOS 10.5 or 11.0) and remove the golden images for other environments including the Ubuntus. I've got ready running docker images for Ubuntus myself, it'll be way faster if I create the reference images.

The CI test automation will pass with missing golden images as well.

sandro:
Great, thank you @tonimelisma for offering to regenerate the golden images for me. Sorry I couldn't be more helpful in that regard. I pushed the changes, let me know if I need to fix anything else.

Thanks!

PR #174, opened 2021-02-22
newgoldens

tonimelisma:


tonimelisma:
Removing macOS-11.0 from Github CI temporarily as it's not yet officially supported and there's lots of issues

PR #175, opened 2021-02-22
Revert macos-10 goldens

tonimelisma:


tonimelisma:
Looks like libvips has been updated in macOS-10.15 homebrew and the old golden reference images no longer work. I don't have a working macOS-10.15 to recreate new ones so we're out for the time being. Simultaneously Github's macOS-11.0 images aren't yet officially supported, cause false negatives and often stall.

So we'll be without macOS CI until Github gets 11.0 working.

PR #177, opened 2021-03-06
vips_find_trim implemented

Andrewsville:
Hello!

In this PR I'm trying to add the very useful [`find_trim` function](https://libvips.github.io/libvips/API/current/libvips-arithmetic.html#vips-find-trim).

Unlike other methods on `ImageRef`, it does not modify the image, it just returns the coordinates (left, top, width, height). In case of an error, it returns -1 for all coordinates.


tonimelisma:
Hey @Andrewsville! Thanks so much for contributing and apologies for the silence. Your PR just slipped right past me. I've got a whirlwind right now at home, we had a burst pipe and been moving out and in on short notice, lots of hassle.

But reviewing the PR it's great code! Unit tests and everything. Thanks so much man!

PR #180, opened 2021-03-30
Fix InterpretationGB16 definition

coffeemakr:
`InterpretationGB16` is set to `C.VIPS_INTERPRETATION_RGB16` instead of `C.VIPS_INTERPRETATION_GB16`


coffeemakr:
Oh no, there is no `VIPS_INTERPRETATION_GB16`

PR #181, opened 2021-03-30
Remove InterpretationGB16

coffeemakr:
The  `InterpretationGB16` definition is a typo of `InterpretationRGB16`.


tonimelisma:
Hey @coffeemakr good catch but it seems like C.VIPS_INTERPRETATION_RGB16 is indeed declared in https://libvips.github.io/libvips/API/current/VipsImage.html under `enum VipsInterpretation`

Perhaps your previous PR was correct?

coffeemakr:
Yes `VIPS_INTERPRETATION_RGB16` is correct, but `InterpretationRGB16` is already defined in `color.go`

PR #184, opened 2021-04-04
Support image_insert()

MaxPower15:
I was porting some code that relied on ruby-vips, and it used the `insert()` operation. When I saw that wasn't available in govips, I tried using `Composite2()` with the ATOP blend mode, and that kind of worked, though I suspect not exactly, and it was about ~30%~ 80% slower than using `insert()`. So I decided to add support for `insert()`!

These changes worked for me, but I'm not totally sure I've followed all the appropriate patterns. Let me know if there are pieces I should change and I'll get it updated.

### DIFF
--- a/vips/conversion.c
+++ b/vips/conversion.c
@@ -148,3 +148,14 @@ int composite2_image(VipsImage *base, VipsImage *overlay, VipsImage **out,
                      int mode, gint x, gint y) {
   return vips_composite2(base, overlay, out, mode, "x", x, "y", y, NULL);
 }
+
+int insert_image(VipsImage *main, VipsImage *sub, VipsImage **out, int x, int y, int expand, int background) {
+  VipsArrayDouble *vipsBackground;
+  double r = (background & 0xff0000) >> 16;
+  double g = (background & 0x00ff00) >> 8;
+  double b = (background & 0x0000ff) >> 0;
+  double background_arr[3] = {r, g, b};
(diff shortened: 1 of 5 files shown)


tonimelisma:
Hey @MaxPower15, thank you so much for the contribution and welcome to the Govips community!

Your PR looks great. I see you added the golden image test, that's perfect. The only thing I'd comment on at this stage are the optional arguments. We don't really have a proper solution to that. If you're interested in the history, reading the comments on the latest issues between David Byttow and me kind of clarifies. We're struggling with how to present the optionals for image loading and exporting as well. There's no clean solution due to Go's lack of null safety. I think this also means that the optional arguments struct you're using could be ambiguous? Perhaps the user only fills one argument in the struct, how will you know that's the case?

Is there any way we could drop the optional arguments, and just e.g. make the mandatory? Or would it be possible to supply nil if the argument is irrelevant in context or the user doesn't want to, and nil would be interpreted as "default"?

In any case, thank you so much for the contribution. IT looks good to me and I'm ready to merge after we have consensus on the optional arguments.

Have a great week!

MaxPower15:
Sure thing, I'll get those changes made. In this case, I think I was fortunate that the defaults for these _particular_ optional arguments matched go's empty defaults... at least as long as the base image doesn't have transparency, heh. Will pursue required args with nil values meaning "use the default." Thanks for the review!

tonimelisma:
Thank you @MaxPower15! Merged! Hope you contribute in the future as well!

Issue #186, opened 2021-04-08
AVIF support


Andrewsville:
Govips IMO currently supports exporting to JPEG, PNG, WebP, HEIF and TIFF. I would really like to see AVIF support.

From the outside it looks pretty straightforward. The go/c part will be easy, ppa:tonimelisma/ppa contains libaom so AVIF support should already be present in CI, at least in the Linux part. I have no idea about the MacOS part, but we'll see when we get there :)

Unless someone is already working on it, I should be able to prepare a PR in a week or two.

tonimelisma:
Hey @Andrewsville! Thanks for the great suggestion. I agree wholeheartedly, AVIF support would be great. I checked the macOS homebrew version of libvips and it seems to support AVIF via libheif if I understand correctly.

In any case a PR would be greatly welcome! Looking forward to the contribution. Thanks again!

PR #190, opened 2021-04-25
AVIF support

Andrewsville:
As promised in #186, here is my attempt at supporting AVIF :)

Pretty much everything is a copy paste from HEIF support with a couple of differences:

* The `isHeif` function was returning TRUE for AVIF as well and this should stay, since HEIF is the container for both HEIC and AVIF. On top of that there is a new `isAvif` function.
* AVIF is exported by the `heifsave` function in libvips. That's why the `ImageTypes` map contains `heif` for AVIF.
* libvips docs say the default quality value is 50 but `NewHeifExportParams` is setting 80, so I used 80.
* In `foreign.c` when loading a heif image, it gets autorotated. Honestly, I have no idea if it should be done for AVIFs too :)

There are also two very basic tests.

So it should be more or less working :) Please take a look and I'll be happy to fix any bug you may find.


tonimelisma:
Hey @Andrewsville and thanks for the PR - looks great. The automated testing is failing on Ubuntu 20.04 for some reason. Could you check the output of the CI and debug?

Andrewsville:
Interesting, tooks like the 20.04 libvips-dev doesn't know heif or something like that. I have the same issue when I try to run the tests locally in a Docker image. I'll investigate it.

tonimelisma:
Hey @Andrewsville did you get any clarity on this?

Andrewsville:
Sorry for the late reply. I finally had the time to investigate and the reason is quite simple. The `ppa:tonimelisma/ppa` repo does not contain libheif for 20.04 meaning the distribution will be used. And its older but more importantly not compiled against libaom so ti doesn't support AVIF.

It have tried to add the PPA repo from libheif authors that contains versions for both 18.04 and 20.04 with all dependencies. Their more modern version requires a newer libde265 than in the distribution, so I have added their PPA of libde265, too.

Let's see if it helped :)

Andrewsville:
And it looks like the golden tests fail with newer deps, especially libheif. Apart from golden tests, everything passes. Could you publish a libheif package for 20.04?

PR #191, opened 2021-05-02
Added getPoint method.

rcilia:


vips/arithmetic.go
```diff
@@ -99,3 +100,16 @@ func vipsFindTrim(in *C.VipsImage, threshold float64, backgroundColor *Color) (i
...
+func vipsGetPoint(in *C.VipsImage, n int, x int, y int) ([]float64, error) {
+	incOpCounter("getpoint")
+	var out *C.double
+	defer gFreePointer(unsafe.Pointer(out))
+
+	if err := C.getpoint(in, &out, C.int(n), C.int(x), C.int(y)); err != 0 {
+		return []float64{}, handleVipsError()
+	}
+
+	return (*[1 << 30]float64)(unsafe.Pointer(out))[:n:n], nil
```
tonimelisma, line 114: Thanks a lot for the PR @rcilia! Great to have contributions from you.

Now I just need to say up front that I'm not enough of an ``unsafe`` guru to review this part over here. Looks like pointer arithmetic 🤯 @davidbyttow can you check this out? Otherwise the PR looks good to merge to me

vips/arithmetic.go
```diff
@@ -99,3 +100,16 @@ func vipsFindTrim(in *C.VipsImage, threshold float64, backgroundColor *Color) (i
 
 	return int(left), int(top), int(width), int(height), nil
 }
+
+// https://libvips.github.io/libvips/API/current/libvips-arithmetic.html#vips-getpoint
+func vipsGetPoint(in *C.VipsImage, n int, x int, y int) ([]float64, error) {
```
davidbyttow, line 105: What I don't understand is why `n` is being passed in here. Isn't that an output value from vips to designate the length of the output vector? I don't think that needs to be referenced from Go, and only used to create the Go array of the appropriate size.

vips/arithmetic.go
davidbyttow: Hmm, yeah this seems a bit unreadable to me. Any way to make this more clear? Also, see my comment about passing in `n`, seems like we shouldn't do that.

davidbyttow (changes requested):

tonimelisma:
Apologies, but I'm closing this PR due to lack of activity. If by any chance you can help out by finishing it, all assistance is greatly welcome!

PR #200, opened 2021-06-15
Add respect `fail` parameter to `pngload`

bozaro:
`pngload` actually has `fail` option support, although this is not reflected in the documentation.


tonimelisma:
Thank you for the active contribution @bozaro!

PR #222, opened 2021-11-20
Support JPEG2000

Elad-Laufer:


Elad-Laufer:
Requires libvips version >8.11.x

PR #224, opened 2021-11-23
Add simple (and naive) implementation of pixelate filter

Elad-Laufer:
A pixelate filter based on simple resize steps.


vips/image.go
```diff
@@ -1180,6 +1180,40 @@ func (r *ImageRef) ModulateHSV(brightness, saturation float64, hue int) error {
 	return nil
 }
 
+// Pixelate applies a simple pixelate filter to the image
+func (r *ImageRef) Pixelate(factor float64) error {
```
davidbyttow, line 1184: do you want to assert that `factor > 0`?

davidbyttow: I feel like this is a little specific to hang on `ImageRef` type. 

How would you feel about making this a utility method that takes in (and maybe return a new ImageRef)? Or it can modify in place

Elad-Laufer: The reason I placed under ImageRef is that to my understanding the same effect can be archived by with a single pass on the image, and I didn't want to end up with two APIs that do the same.

davidbyttow: Sorry, all I'm saying is that we could easily do:

```
func Pixelate(image *ImageRef, factor float64) error {
  // your code
}
```

I mean, I know there's a bunch of things on ImageRef already, I do wish it were more minimal and the "modifiers" of it were separate functions. WDYT? Let me know and we can merge

Elad-Laufer: ok, this way we won't end up with BC later on.

davidbyttow: BC?

Elad-Laufer: Backwards Compatible :)

Elad-Laufer: > Sorry, all I'm saying is that we could easily do:
> 
> ```
> func Pixelate(image *ImageRef, factor float64) error {
>   // your code
> }
> ```
> 
> I mean, I know there's a bunch of things on ImageRef already, I do wish it were more minimal and the "modifiers" of it were separate functions. WDYT? Let me know and we can merge

Implemented

PR #226, opened 2021-11-28
GetPoint method

cshum:
Follow up from last closed pull request:
- Determine n by whether image has alpha channel.
- It seems maximum size of array is 4, so we need not allocate more than that. 
- I am not a unsafe guru either. But reading from other unsafe implementations, this seems to be pretty common of converting unsafe pointer to array?

I am trying to implement edge trimming with FindTrim, but needing the color to be automatically determined. The GetPoint method fills in the gap.


vips/image.go
```diff
@@ -1235,6 +1235,16 @@ func (r *ImageRef) FindTrim(threshold float64, backgroundColor *Color) (int, int
 	return vipsFindTrim(r.image, threshold, backgroundColor)
 }
 
+// GetPoint reads a single pixel on an image.
+// The pixel values are returned in a slice of length n.
+func (r *ImageRef) GetPoint(x int, y int) ([]float64, error) {
```
davidbyttow, line 1240: why `GetPoint` vs `GetValue` or `GetColor`? I guess it mimicks vips... Just curious if you thought of this

cshum: Maybe GetColor is more of RGB and may not necessary cover alpha channel? GetValue sounds too ambiguous?
Since this is a libvips binding, I guess it would be better name it as close as possible?
Seems others are naming it the same way too https://libvips.github.io/pyvips/vimage.html#pyvips.Image.getpoint

cshum: Maybe GetColor is more of RGB and may not necessary cover alpha channel? GetValue sounds too ambiguous?
Since this is a libvips binding, I guess it would be better name it as close as possible?
Seems others are naming it the same way too https://libvips.github.io/pyvips/vimage.html#pyvips.Image.getpoint

PR #241, opened 2022-01-04
Fix typos and shadowing

Elad-Laufer:


davidbyttow:
Thanks @Elad-Laufer !

PR #242, opened 2022-01-04
Remove deprecated export functions

Elad-Laufer:
ToImage() always passes through PNG as it is guaranteed to be supported and lossless

### DIFF
--- a/examples/logging/example_logging_test.go
+++ b/examples/logging/example_logging_test.go
@@ -62,8 +62,7 @@ func main() {
 	checkError(err)
 	myLogger("loggingExample", vips.LogLevelInfo, fmt.Sprintf("Before: %v x %v", image1.Height(), image1.Width()))
 
-	ep := vips.NewDefaultJPEGExportParams()
-	image1bytes, _, err := image1.Export(ep)
+	image1bytes, _, err := image1.ExportJpeg(nil)
 	checkError(err)
 	err = ioutil.WriteFile("output.jpg", image1bytes, 0644)
 	checkError(err)
--- a/examples/tiff/example_tiff_image.go
+++ b/examples/tiff/example_tiff_image.go
@@ -30,10 +30,9 @@ func main() {
 	checkError(err)
 	fmt.Println(inputImage.Height(), inputImage.Width())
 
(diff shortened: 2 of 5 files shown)


tonimelisma:
Is this backwards compatible?

Elad-Laufer:
> Is this backwards compatible?

No. I've removed exported functions.

davidbyttow:
Thanks @Elad-Laufer !

tonimelisma:
Hey @davidbyttow are you sure we want to break backwards compatibility? Or does this mean we should bump up the major version number?

davidbyttow:
Hey @tonimelisma to be honest I didn’t give this too much deep thought. My thinking is that people should be pinning to a specific version, but then again I know that’s not the reality. Perhaps we should revert this and then bump the major (or minor since this has been deprecated for a while now) and then roll forward. Though if we are going to bump the major version then I would like to see if there are other changes to fold in. 

WDYT @tonimelisma @Elad-Laufer ?

tonimelisma:
I've tried to follow Go's [backwards compatibility policy](https://go.dev/blog/module-compatibility) pretty strictly. So basically bump up the major version for breaking changes.

I've tried to think about other changes we would want to do simultaneously with a version bump. I wonder where are with regards to supporting all libvips importing and exporting parameters? Would there be breaking changes there we could implement simultaneously?

Might this also be a good time to refactor the export parameters to an alternative that's easier for both us as well as govips users? With the struct we use there's the challenge that we can't see when a user overrides any specific parameters on the govips side. For users, it's a bit more complicated to always get a struct from ``NewDefaultJPEGExportParams()``, fill it in and then pass it back, instead of a simpler variadic interface (such as e.g. [here](https://pkg.go.dev/github.com/disintegration/imaging#Open)). Not sure if such a change would be worth the effort.

davidbyttow:
Good points, will consider though, but generally agree on the refactor. I need to get my head back into this for a day to form an opinion.

tonimelisma:
Sure. Apologies for the lack of availability on my end recently and for the coming weeks, I'm just relocating to SF with my family and there's a lot on my table.

Issue #117, opened 2020-11-09
Support animated images


Elad-Laufer:
Test and if _needed_ implement the following for animated images:

Note: libvips loads the pages into a roll

**IO**

- Load all image pages
- When saving to formats that do not support animation (such as JPG) - save the first page and not the roll

**Operations**

1. Resize - might need to adjust the page hight property
2. Rotate - might need to splice a new roll
3. Flip - make sure that the page order does change
4. ExtractArea - might need to splice a new roll
5. Label - label a selected page (all pages as an option)?
6. Extend - might need to splice a new roll

**Discussion**

Composite:

compose on each page - I think this should be the default behavior, as we use compose primarily for watermarking
compose 2 animated images? - should we define the behavior?

tonimelisma:
Looks great to me. Likely would be most logical if Label() and Composite() both function in a similar fashion. I think the primary use case of both is watermarking so probably makes sense if both default to all pages.

Elad-Laufer:
https://github.com/davidbyttow/govips/pull/246

AttilaTheFun:
@Elad-Laufer this seems to be working now! Do you want to close it? Or is there still work to be done.

Elad-Laufer:
> @Elad-Laufer this seems to be working now! Do you want to close it? Or is there still work to be done.

Not all the features are supported yet, currently only resize is covered 100% for multipages images.

See https://github.com/davidbyttow/govips/issues/248 and https://github.com/davidbyttow/govips/issues/249 for example.

davidbyttow:
We should perhaps re-word this or close the issue and just use other issues for the additional features. We technically now support animated images.

davidbyttow:
Closing this for now

PR #246, opened 2022-01-13
Animated GIF support

Elad-Laufer:
- Export native of GIF
- Support animated images resize
- Support animated images rotate
- Load from file with import options


Elad-Laufer:
https://github.com/davidbyttow/govips/issues/117

davidbyttow:
Resolve conflicts?

AttilaTheFun:
I'm super excited for this to land! Right now I have some "if-gif" code to fall back on the builtin image package when dealing with gif images that I would love to delete.

AttilaTheFun:
Awesome! @davidbyttow could we cut a new release including this commit?

davidbyttow:
Done! https://github.com/davidbyttow/govips/releases/tag/v2.10.0 @AttilaTheFun

AttilaTheFun:
Sweet I'll try it out later today. If I recall before, gifs would break when I tried to resize them. 

As a result I had to store gifs in a separate S3 bucket with a separate CloudFront bucket which had my Lambda@Edge function disabled.

If gif resizing is working now, I'd be excited to unify gifs with the rest of the images on my server.

AttilaTheFun:
@davidbyttow @Elad-Laufer So I just pulled down this version and tested it out but I'm still seeing the same issue I saw before. If I import a GIF from a file and then export it back to a file, it only seems to write the first frame of the GIF.

I'm not sure if this is still broken or if I'm "holding it wrong", but I've attached the images I used to test:
| Original | VIPS |
| - | - |
sRGB IEC61966-2 1 | sRGB IEC61966-2 1 Encoded |

I'm importing the image from a file handle like so:
```
	img, err := vips.NewImageFromReader(reader)
	if err != nil {
		return nil, err
	}
```

And then I'm exporting it this way:
```
		// Export a GIF:
		ep := vips.NewGifExportParams()
		bytes, _, err := img.ExportGIF(ep)
		if err != nil {
			return nil, err
		}
```

And writing the bytes back to the new file. Is this supposed to be working? Or is there another way to import / export GIFs I should be using?

Elad-Laufer:
> That said, I believe defaulting to only loading the first frame of a GIF is a bad UX. I'd advocate for making the default value be -1 so that importing and exporting an image with the default parameters would round trip.

I agree, but that would change the current behavior and break backwards compatibility in an unexpected way as it will compile just fine and will behave differently. For example, when converting an animated image to a format that does not support animation such as JPEG, if the more than one page is loaded the exported image will the entire roll and not a single frame (which defaults to the first one).

davidbyttow:
There could be `NewGifImportParams()`?

PR #262, opened 2022-02-03
Set and Get PageDelay for animated GIF

cshum:
Set and Get page delay for animated GIF:
- image.SetPageDelay
- image.PageDelay


vips/image.go
```diff
@@ -716,6 +716,25 @@ func (r *ImageRef) SetPageHeight(height int) error {
...
+	if n <= 1 {
+		// should not call if not multi page
+		return nil, nil
+	}
+	return vipsImageGetDelay(r.image, n)
+}
+
+// SetPageDelay set the page delay array for animation
+func (r *ImageRef) SetPageDelay(delay []int) error {
+	var data []C.int
```
davidbyttow, line 731: should this also early return if n pages > 1

cshum: https://www.libvips.org/API/current/libvips-header.html#vips-image-get-n-pages 
the `vips_image_get_n_pages` only represents the number of pages of the original image, not the number of pages loaded/available to process. 

And it is entirely possible to construct an animation out of a static image (where pages return 1), by manually constructing the delay array, with the corresponding page grid.

cshum: The n checking on the Get function only exists for C array scanning (user should not input n manually). Though I am not sure if that is valid in every cases...

cshum: I just added another other test https://github.com/davidbyttow/govips/pull/262/files#diff-d5564bc69d6e9fa157b1792e11fee1ff846d38dd2d18bf25019afad1aa87ce11R121
even though I loaded only 3 pages, the delay array still shows a valid array int of size 8: `[100, 100, 100, 100, 100, 100, 100, 100]`.

So the the `delay` array works the same way as `n` - it only represent the original metadata, does not change not matter how it is loaded.

But when I set the size 8 array to the 3 frames animation, it still produces a valid result

Issue #255, opened 2022-01-22
Resizing GIFs is failing on linux, hanging / breaking on macOS for larger GIFs


AttilaTheFun:
Hi @Elad-Laufer and @davidbyttow! I wanted to try out the GIF resizing on my server in a Go container-based lambda function. 

Unfortunately, while I was able to resize a GIF locally with the latest version of govips, when I tried to resize it in my lambda function I got this error:

```
"VipsOperation: class \"magicksave_buffer\" not found\n\n
Stack:\ngoroutine 1 [running]:\nruntime/debug.Stack()\n\t/usr/local/go/src/runtime/debug/stack.go:24 
+0x65\ngithub.com/davidbyttow/govips/v2/vips.handleVipsError()\n\t/go/pkg/mod/github.com/davidbyttow/govips/v2@v2.10.0/vips/error.go:39 
+0x57\ngithub.com/davidbyttow/govips/v2/vips.handleSaveBufferError(0xc000044000)\n\t/go/pkg/mod/github.com/davidbyttow/govips/v2@v2.10.0/vips/error.go:32 +0x25\ngithub.com/davidbyttow/govips/v2/vips.vipsSaveToBuffer({0x7ffbacea87f0, 0x0, 0x5, 0x0, 0x0, 0x4b, 0x0, 0x0, 0x1, 0x0, ...})\n\t/go/pkg/mod/github.com/davidbyttow/govips/v2@v2.10.0/vips/foreign.go:437 
+0xa5\ngithub.com/davidbyttow/govips/v2/vips.vipsSaveGIFToBuffer(0x7ffbacea87f0, {0x0, 0x0})\n\t/go/pkg/mod/github.com/davidbyttow/govips/v2@v2.10.0/vips/foreign.go:432 
+0xf8\ngithub.com/davidbyttow/govips/v2/vips.(*ImageRef).ExportGIF(0xc000244050, 0xc2)\n\t/go/pkg/mod/github.com/davidbyttow/govips/v2@v2.10.0/vips/image.go:860 
+0x59\ngithub.com/attilathefun/utils/imageutil.(*Image).EncodeToBytes(...
```

From the name of the missing class, I assumed that govips might be depending on imagemagick so I installed that but still got the same error.

Do you have any idea what could be causing this?

My Dockerfile is as follows:
```
###############################################################################
# STEP 1: Build Executable ####################################################
###############################################################################

FROM golang:1.17-alpine as builder

# Install packages:
[...]

PR #269, opened 2022-02-16
Default to using gifsave for libvips >= 8.12.0

vansante:
I use libvips without magick and I would like to start generating gif files, which is possible since [libvips version 8.12.0](https://www.libvips.org/2021/11/14/What's-new-in-8.12.html). However `govips` still hardcodes the codepath to use `magicksave` for all gif exports. It results in the error `VipsOperation: class "magicksave_buffer" not found` (see https://github.com/davidbyttow/govips/issues/255 as well).

This PR does a version detection to check whether the libvips version `>= 8.12.0` and if true, defaults to using the new gifsave function which is much faster.

This is technically backwards incompatible (you need a new dependency [CGIF](https://github.com/dloebl/cgif), the ubuntu package name is `libcgif-dev`) otherwise the GIF saving feature is broken. I tried doing a runtime feature detection check for cgif, but couldn't find a way to make it work. However, currently trying to save gifs without having magick also results in an error, so I'm not sure how much of a problem this is.

Let me know if I need to tweak this PR further :).

### DIFF
--- a/vips/foreign.c
+++ b/vips/foreign.c
@@ -329,9 +329,17 @@ int set_magicksave_options(VipsOperation *operation, SaveParams *params) {
 
 // https://libvips.github.io/libvips/API/current/VipsForeignSave.html#vips-gifsave-buffer
 int set_gifsave_options(VipsOperation *operation, SaveParams *params) {
-  // TODO: Set optional arguments?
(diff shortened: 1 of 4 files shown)


davidbyttow:
Thank you!

PR #273, opened 2022-03-28
Support restart interval params in govips

sbraty:
Hello, I use libvips to process images and I have huge pipline of processing.

I would like to optimize this pipline and found out that libvips start supporting restart markers ([libvips changelog](https://www.libvips.org/2021/11/14/What's-new-in-8.12.html), they called it "restart interval"), that helps to faster process images in next steps.

If need some tweak, feel free to ping me.


sbraty:
@davidbyttow can you please take a look?

vips/foreign.go
```diff
@@ -332,6 +332,10 @@ func vipsSaveJPEGToBuffer(in *C.VipsImage, params JpegExportParams) ([]byte, err
 	p.jpegOptimizeScans = C.int(boolToInt(params.OptimizeScans))
 	p.jpegQuantTable = C.int(params.QuantTable)
 
+	if MajorVersion == 8 && MinorVersion >= 12 {
```
tonimelisma, line 335: Should this also work for major version 9 and later, when they will eventually come?

tonimelisma (commented):
Hey @sbraty thank you very much for the contribution and apologies it's taken so long to review. There's one small change for the version control I was wondering about, and could you also check unit tests?

tonimelisma:
Hey @sbraty any updates on this, either?

PR #278, opened 2022-04-08
Add header getters

sbraty:


tonimelisma:
Thanks @sbraty for the contribution! I didn't see any unit tests, could you add them? Also, can you clarify what sort of use cases you think these will be useful for?

tonimelisma:
Hey @sbraty  any update on unit tests, and possibly updating your branch so it's compatible with the latest on main?

PR #284, opened 2022-05-25
Add `PngExportParams.Filter` option to .png export

bozaro:
I want to implement something like `optipng`: try compress .png with differ filter and get smallest image.

More info about the PNG delta filters can be found at http://optipng.sourceforge.net/pngtech/optipng.html on section 2.2.


bozaro:
Interesting observation: looks like `C.VIPS_FOREIGN_PNG_FILTER_NONE` is default filter value for .png.
But `0` is act as `auto` and select best filter on my test images...

PR #299, opened 2022-08-03
Original image type

Elad-Laufer:
When implicitly converting the image type (BMP to PNG), keep the original format available for query


tonimelisma (approved):
lgtm

PR #307, opened 2022-09-27
gaussian blur support min_ampl

jishi92:


tonimelisma:
Hey @jishi92 thanks for the contribution! Could you write unit tests as per the CONTRIBUTING document? Just some simple passing and failing test to check the parameter, perhaps with the golden images? There's plenty of examples in the code.

Thank you!!

jishi92:
Hi @tonimelisma,  I create a issue [#309] and write a unit test according to the CONTRIBUTING document.
I change the `jpg-24bit.GaussianBlur-macos-12.golden.jpeg`, because the origin one is wrong .
I  just need to make sure that the old parameters in GaussianBlur are the same as the old result.

tonimelisma (approved):
lgtm. Should really clean up the golden unit tests and get some system to autoupdate the golden images when libvips announces a new version

PR #317, opened 2022-10-24
Add APIs to get image EXIF data and ICC profile data

AttilaTheFun:
These are two APIs I've wanted for a while and I finally got around to implementing them + making a pull request.

The first, GetICCProfile, is related to HasICCProfile but it actually returns the byte slice resulting from invoking vips_image_get_blob with the VIPS_META_ICC_NAME parameter. 

This is useful for me as well as hopefully other consumers because I want to parse the data with LCMS2 and determine the ICC profile name and whether it's a wide color profile or not.

The second API, GetExif, is related to ImageFields but it actually returns the string values in a map for all of the image fields with an "exif" prefix.

This is useful for pulling interesting information out of images like the type of camera and lens.

### DIFF
--- a/vips/header.c
+++ b/vips/header.c
@@ -6,6 +6,19 @@ unsigned long has_icc_profile(VipsImage *in) {
   return vips_image_get_typeof(in, VIPS_META_ICC_NAME);
 }
 
+unsigned long get_icc_profile(VipsImage *in, const void **data,
+                              size_t *dataLength) {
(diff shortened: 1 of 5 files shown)


AttilaTheFun:
@tonimelisma @davidbyttow Hey folks! I remember we talked about this a long time ago and I finally got around to implementing it. Let me know what you think!

tonimelisma (approved):
lgtm

AttilaTheFun:
Awesome, thanks!

Issue #313, opened 2022-10-09
Could we add a Bitdepth parameter to HIEF + AVIF export params?


AttilaTheFun:
With the release of iOS 16 and soon macOS Ventura, Apple platforms will finally support AVIF. 
(AV1 support for videos is still not available yet sadly.) See: https://caniuse.com/avif

I'm interested in using AVIF for high bit depth images on these devices. In December last year, libvips added support for 8/10/12 bit HEIC and AVIF load and save: https://github.com/libvips/libvips/pull/2596

However, AFAICT neither of the govips versions of the export parameters take the bitdepth argument. Would it be possible to add support for this to govips? I could create a PR if I could get some code pointers. I can see the export params structs but I'm not entirely sure what else needs to happen to pass that to the underlying libvips.

AttilaTheFun:
My PR fixed this issue.

PR #322, opened 2022-10-28
Add support for higher bit depth AVIF and HEIF images

AttilaTheFun:
Back in February, John helped me add support for loading and saving 8/10/12 bit HEIC and AVIF files to libvips here:
https://github.com/libvips/libvips/pull/2596

Now with the release of iOS 16 / MacOS Ventura which support AVIF images natively, I want to add support for them to my app. 

I asked about this a while ago here: https://github.com/davidbyttow/govips/issues/313 and finally had some time to implement it myself.

The changes in this PR are as follows:
- I cleaned up the HEIF / AVIF export parameters so they both now take a Bitdepth parameter like PNGs. Also libvips has renamed "speed" to "effort" so I migrated us to use that parameter name with backwards support for the speed parameter for code that used it.
- I added golden image tests for 8bit AVIF and encoding a 16bit TIF to a 12bit AVIF.
- The files I touched got auto formatted by gofmt so we should probably just commit those changes to avoid annoying future PR authors.


AttilaTheFun:
Hi again @tonimelisma and @davidbyttow 😊
I just updated to iOS 16 and MacOS Ventura and I was excited to add AVIF support to my app.
Let me know what you think!

AttilaTheFun:
Ah I see this failed the CI job because of the error:
heifsave_buffer: no property named `bitdepth'

I suspect this is because the builders have libvips 8.12.X or earlier while this param was added in 8.13.X which I have locally. I'll add a check for the libvips version.

tonimelisma:
Hey @AttilaTheFun it's difficult to read the PR as you've used a different formatter for the code. It looks like 90% of the changes are just line breaks and spacing? Would it be possible to remove these changes and just show the actual code changes?

AttilaTheFun:
Hey Tony! Sure thing -- I can do that after dinner.

I use VSCode with the default Go tools installed, though. It might be worth running gofmt once on the entire codebase for one commit rather having people fighting the formatter on every pr.

I think there is a way to create a gofmt config that is checked into the repo if you don't like the default settings.

tonimelisma:
I think all Go formatting is the same, unless you're adding line breaks manually to existing code. It's probably the C code formatting which you have different settings for.

tonimelisma (approved):
lgtm

PR #323, opened 2022-11-01
Fix bug with icc transform bit depth

AttilaTheFun:
This PR fixes a bug with the TransformICCProfile API. Here I was passing 0 for the bit depth (leaving it unset) but the default value if you do this is 8-bit. This causes ICC transforms of >8-bit images to revert to 8-bit and become distorted. The code I'm adding here was copied from the adjacent OptimizeICCProfile method which defaults to 16-bit processing unless the source image was already 8-bit.


AttilaTheFun:
@tonimelisma This one is a pretty straightforward bug fix 😅

tonimelisma (approved):
lgtm

Issue #334, opened 2022-11-23
Getting object alive count greater than 0 when calling vips to generate a black image


abhilather:
We are using this benchmarking code through govips for basic operation of creating black images (https://pkg.go.dev/github.com/davidbyttow/govips/v2/vips#Black).

```
func BenchmarkVipsBlack(b *testing.B) {
	for n := 0; n < b.N; n++ {
		_, _ = vips.Black(100, 100)
	}
	vips.PrintCache()
	vips.PrintObjectReport("BenchmarkVipsBlack")
	vipsMemStats := &vips.MemoryStats{}
	vipsRuntimeStats := &vips.RuntimeStats{}
	vips.ReadVipsMemStats(vipsMemStats)
	vips.ReadRuntimeStats(vipsRuntimeStats)
	golog.Logger.Info(fmt.Sprintf("VIPS Runtime Stats: %+v", vipsRuntimeStats))
	golog.Logger.Info(fmt.Sprintf("VIPS Memory Stats: %+v", vipsMemStats))
}
```
The response for this as mentioned below -

```
vips live objects: BenchmarkVipsBlack...
2 objects alive:
0) VipsBlack (0x12d840e10), count=1
VipsBlack (black), make a black image, black out=((VipsImage*) 0x13900b000) width=100 height=100 -
1) VipsImage (0x13900b000), count=1
VipsImage (image), image class, 100x100 uchar, 1 band, multiband
2022/11/23 18:44:22 [govips.info] 
2022-11-23T18:44:22.566+0530    INFO    VIPS Runtime Stats: &{OperationCounts:map[]}
2022-11-23T18:44:22.566+0530    INFO    VIPS Memory Stats: &{Mem:0 MemHigh:0 Files:0 Allocs:0}
goos: darwin
goarch: amd64
pkg: irs/imagerendering
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkVipsBlack-12           Operation cache:
2022/11/23 18:44:22 [govips.info] 
=======================================
vips live objects: BenchmarkVipsBlack...
22 objects alive:
0) VipsBlack (0x139009540), count=1
VipsBlack (black), make a black image, black out=((VipsImage*) 0x13900b7d0) width=100 height=100 -
1) VipsBlack (0x10500e100), count=1
VipsBlack (black), make a black image, black out=((VipsImage*) 0x13900e340) width=100 height=100 -
2) VipsBlack (0x10500e000), count=1
[...]

PR #361, opened 2023-05-30
Added missing finalizer method

rektabhi:
In a few methods in `image.go` the default constructor for creating an ImageRef object was getting used.

```
 &ImageRef{image: out}
```
The problem with this is that the `finalizer` is not set on these objects and thus `Close()` is never called. This keeps on filling the memory with zombie objects.

Similar issue listed here: https://github.com/davidbyttow/govips/issues/334 and https://github.com/davidbyttow/govips/pull/336


tonimelisma:
Hey @rektabhi why did you close this PR?

rektabhi:
@tonimelisma Some tests were failing, I will re-raise after fixing the tests. Did not get time to look at the failing tests.

PR #362, opened 2023-05-30
add windows support

lukeparkf12:
add windows support using static build dll file and header files of libvips.
all dll files from https://github.com/libvips/build-win64-mxe/releases/download/v8.14.2/vips-dev-w64-web-8.14.2-static.zip, and should be copied into PATH folder of your system to load dll files on execution.


tonimelisma:
Thanks @lukeparkf12! This is great stuff!

It looks like you're bundling the DLL files into the govips repository. I'm not sure if that's the best way to go around this. Perhaps you should just include the instructions in the README on how to get and install the dependencies from an official source.

Is your Windows port generally strict about the libvips version? Do you think you'll have time to maintain it?

Do you think you could also change the Github Actions CI to include a Windows-run? That would ensure that the Windows parts are up to date

lukeparkf12:
I am sorry I do not have a time for maintainning.
What I did were just adding latest include and libs from official site and platform go file for cgo config to pass build because I cannot find a way to use system PATH for LDFLAGS.
I think it should be closed without merging because PR was not enough.
I hope you know it could be easy wat just bundling includes and libs for windows when we have reliable official distributor of dll and headers.
Thank you.

tonimelisma:
Thanks Luke!

PR #370, opened 2023-07-20
Streaming support

negator:
Adds Source and Target for full streaming support.

https://www.libvips.org/2019/11/29/True-streaming-for-libvips.html

This is a fairly straightforward patch. We add two new functions `NewSourceFromReader` and `NewTargetToWriter`, along with an `ExportWriter` function and all secondary `Export[Jpeg|Tiff...]Target` functions. 

Inspired by: https://github.com/vansante/go-vips-thumbnailer

cc @tonimelisma


vips/foreign.go
```diff
@@ -131,6 +132,20 @@ func IsTypeSupported(imageType ImageType) bool {
 	return supportedImageTypes[imageType]
 }
 
+// DetermineImageTypeFromSource attempts to determine the image type of the given source
+func DetermineImageTypeFromSource(source *Source) ImageType {
+	length := 20
+	data := C.source_sniff(source.vipsSrc, C.ulong(length))
+	sh := &reflect.SliceHeader{
```
krestovolt, line 139: Hi @negator, I think you need to add `nil` check here, something like this
```go
if data == nil {
		return ImageTypeUnknown
}
```
I got this error when running test `TestNewImageFromReaderFail` (vips/image_test.go:913):
```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x40abb4]
```

negator: Great find thank you!

negator:
ping @AttilaTheFun

AttilaTheFun:
Hey @negator ! I've contributed to the repo previously but I don't have permission to merge PRs. cc @tonimelisma @davidbyttow

davidbyttow:
This looks great! Give me a day to get back to this and respond.

davidbyttow:
I'd love to see more tests added, but will merge.

davidbyttow (approved):

davidbyttow:
noticed BMP failure with "not supported", not sure if related?

negator:
Interesting, let me take a look at get back you.

negator:
@davidbyttow I added several more tests to ensure all image types can be read and written. As well as updated the golden tests to ensure exporting to writers works correctly. 

Some things I discovered:
- Writing HEIC/HEIF files to a closed writer will segfault (using latest `vips-8.14.3`). And there isn't really a good way to determine if a writer closed prior to writing to it. We can exclude these files from streaming writes if it seems prudent.
- TIFF files require the writer to also be a Reader and Seeker (io.ReadSeeker). I've added code to disallow use of a non ReadSeeker when exporting TIFF to a writer. https://www.libvips.org/API/current/VipsTargetCustom.html#vips-target-read

negator:
I discovered there are severe performance penalties that are incurred when C code calls Go code, which this implementation requires. Especially since it performs many 100s of C-to-Go calls from both the reader and writer. I don't really feel comfortable merging this as-is since it might affect performance for the users of the NewImageFromReader function. Additionally the ExportWriter functions will differ similar performance issues.

However I believe there will be a performance boost for users of the NewImageFromFile function which leverages the input streaming capability of libvips from files. Replacing the ExportWriter functions with functions that export to files will similarly yield a more performant implementation.

For now I'd like to close this PR and potentially reopen one with the above mentioned updates at a later time. Let me know how your like to proceed.

PR #409, opened 2024-02-25
Update DevOps

tonimelisma:


tonimelisma:
Hey @davidbyttow, can you please review and approve this? I think you turned on mandatory reviews for merging and I can no longer push stuff personally without your help. Was that intentional?

I'm dropping macos from the test automation because the constantly changing homebrew libvips version is breaking the tests and making image reference tests useless (they're always passing).

I'm also personally getting issues with running the tests due to HEIF saving issues. If you can check if you get issues with running the tests that would be awesome.

davidbyttow (approved):

davidbyttow:
Not intentional. Will fix.  Thank you for this.

PR #414, opened 2024-03-05
Imrove golden image tests

toaster:
This PR contains some improvements which I made when working on a fuzzy image test.
It does not affect any logic or test outcome.

But it changes the test behaviour: missing golden images now let the tests fail. Thus, they are detected by CI runs.

### DIFF
--- a/vips/foreign.go
+++ b/vips/foreign.go
@@ -167,10 +167,10 @@ func DetermineImageType(buf []byte) ImageType {
 	}
 }
 
-var jpeg = []byte("\xFF\xD8\xFF")
+var jpegHeader = []byte("\xFF\xD8\xFF")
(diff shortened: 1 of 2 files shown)


tonimelisma:
Hey, I think you're right that slowing down the testing is not really worth it for the fuzzy matchin. I'm not sure if you realized but not having existing reference images causes creation of them so with this PR it would be impossible to create new reference images.

Thanks for the thoughtfulness but I don't think we want to venture down this path with the test automation.

toaster:
@tonimelisma I would kindly ask you to reconsider your decision.
This PR makes the missing golden images more visible by failing the tests and therefore the CI run but it still creates the missing golden images.

Issue #359, opened 2023-05-22
Quality param ignored on GIFs


wmichelin:
I am working on GIF compression, and the resultant files, even when aggressively using the quality param, are virtually always larger than the source GIF, by a significant margin.

Looking at the VIPs library documentation, I don't see any quality param supported at all for GIFs, which leads me to believe that this parameter may be [erroneously included in the govips bindings](https://github.com/davidbyttow/govips/blob/c6838fceef8d93bca9e044187a412540b62c6090/vips/image.go#L312-L318). 

Please let me know your thoughts. Thank you!

g-mero:
Try this.
```
img.ExportGIF(&vips.GifExportParams{
			StripMetadata: true,
			Quality:       75,
			Effort:   7,
			Bitdepth: 8,
		})
```

n0vad3v:
From `govips` code, I found the Quality or Q is not included in `set_gifsave_options`, so setting `Quality` here might be useless.

```go
// https://libvips.github.io/libvips/API/current/VipsForeignSave.html#vips-gifsave-buffer
int set_gifsave_options(VipsOperation *operation, SaveParams *params) {
  int ret = 0;
  // See for argument values: https://www.libvips.org/API/current/VipsForeignSave.html#vips-gifsave
  if (params->gifDither > 0.0 && params->gifDither <= 1.0) {
    ret = vips_object_set(VIPS_OBJECT(operation), "dither", params->gifDither, NULL);
  }
  if (params->gifEffort >= 1 && params->gifEffort <= 10) {
    ret = vips_object_set(VIPS_OBJECT(operation), "effort", params->gifEffort, NULL);
  }
  if (params->gifBitdepth >= 1 && params->gifBitdepth <= 8) {
      ret = vips_object_set(VIPS_OBJECT(operation), "bitdepth", params->gifBitdepth, NULL);
  }
  return ret;
}
```

From `libvips` there is also no Quality option:

```
/**
 * vips_gifsave: (method)
 * @in: image to save
 * @filename: file to write to
 * @...: %NULL-terminated list of optional named arguments
 *
 * Optional arguments:
 *
 * * @dither: %gdouble, quantisation dithering level
 * * @effort: %gint, quantisation CPU effort
 * * @bitdepth: %gint, number of bits per pixel
 * * @interframe_maxerror: %gdouble, maximum inter-frame error for transparency
 * * @reuse: %gboolean, reuse palette from input
[...]

PR #417, opened 2024-03-18
Refine `GifExportParams`

n0vad3v:
In response to https://github.com/davidbyttow/govips/issues/359#issuecomment-2002981134

Changes:
1. Added notes on `GifExportParams` to tell users that different vips version will use different params
2. `StripMetadata` is removed as `libvips` does not support this function on GIFs, I've tried pass it in with `VIPS_OBJECT(operation), "strip", params->stripMetadata,`, the output image remains the same.
3. Passed in `bitdepth` in early vips version

However this PR is not finished, I've found several problems here.
1. No matter how I change GifExportParams's `Dither`, `Effort` or `Bitdepth`, the output image's sha256sum remains the same
2. In the current `set_gifsave_options`, only one parameter can be used with passed in multiple parameters, as below
```c
  // See for argument values: https://www.libvips.org/API/current/VipsForeignSave.html#vips-gifsave
  if (params->gifDither > 1 && params->gifDither <= 10) {
    ret = vips_object_set(VIPS_OBJECT(operation), "dither", params->gifDither, NULL);
  }
  if (params->gifEffort >= 1 && params->gifEffort <= 10) {
    ret = vips_object_set(VIPS_OBJECT(operation), "effort", params->gifEffort, NULL);
  }
  if (params->gifBitdepth >= 1 && params->gifBitdepth <= 8) {
      ret = vips_object_set(VIPS_OBJECT(operation), "bitdepth", params->gifBitdepth, NULL);
  }
```
But, as different parameters will result in same output image, I'm not sure why this is happening.

### DIFF
--- a/vips/foreign.c
+++ b/vips/foreign.c
@@ -332,7 +332,7 @@ int set_magicksave_options(VipsOperation *operation, SaveParams *params) {
 int set_gifsave_options(VipsOperation *operation, SaveParams *params) {
   int ret = 0;
   // See for argument values: https://www.libvips.org/API/current/VipsForeignSave.html#vips-gifsave
-  if (params->gifDither > 1 && params->gifDither <= 10) {
+  if (params->gifDither > 0.0 && params->gifDither <= 10) {
     ret = vips_object_set(VIPS_OBJECT(operation), "dither", params->gifDither, NULL);
   }
   if (params->gifEffort >= 1 && params->gifEffort <= 10) {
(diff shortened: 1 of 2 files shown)


tonimelisma:
Hey @n0vad3v, thank you very much for the PR. Do you think you will have time finishing it?

n0vad3v:
Hi @tonimelisma I think this PR is ready to merge for now, however more testing might be needed on GIF images.

tonimelisma:
It looks like you've removed the StripMetadata parameter. I'm really sorry but we can't have breaking changes in the API. I know it's not optimal to have poorly working or nonfunctional parameters in the API but we can't break users' software.

n0vad3v:
Got it, now I'm adding `StripMetadata` option back and added a note for it.

tonimelisma:
Thank you sir. Your contributions are always welcome!

PR #433, opened 2024-06-03
Add support for the `vips_gamma()` function via a new func `Gamma()` on `ImageRef`

jawngee:


tonimelisma:
Hey @jawngee! Thanks so much for the PR. Let me know when it's ready for review!

jawngee:
I think it's ready :)

I wasn't really sure about the test though because I'm not sure how to test it.  So all the test does is make sure it doesn't throw an error when called.

Thanks,

Jon.

PR #442, opened 2024-07-05
write icc files to tmp only when necessary

code1super:
I see no reason to write these files to a temporary file when it is not consumed.
It can make unnecessary errors when there is no have temporary folder or when there is no disk space or when there are no write permissions.

### DIFF
--- a/vips/govips.go
+++ b/vips/govips.go
@@ -183,7 +183,9 @@ func Shutdown() {
 		return
 	}
 
-	os.RemoveAll(temporaryDirectory)
+	if temporaryDirectory != "" {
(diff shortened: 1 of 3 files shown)


tonimelisma:
Thanks @code1super and apologies for the delay, I've been traveling overseas. Could you write tests for the changes please?

The PR is greatly appreciated!

tonimelisma:
Hey @code1super I love the PR but could you add the tests?

rhafer:
@tonimelisma @code1super I just created #456 which contains @code1super's original commits plus adjustments to the test. I hope that's fine.

PR #451, opened 2024-10-09
Support PDF files with magic byte in the header

karthik-bashetty:


tonimelisma:
Hey @karthik-bashetty, thanks for this. I'm a bit confused, it looks like your PR adds two golden reference image GIFs? Was that intentional?

Is the only change that you define the first 1024 bytes of the buffer for the PDF? And then you're shuffling the if/else control structure a little bit?

Could you add one of the offending PDFs as an example golden test please? I'd love to merge if you can provide tests for this (even though it looks like a small change).

karthik-bashetty:
Hi @tonimelisma 

Yes the only change is to check the first 1024 bytes of the pdf and restructuring the pdf check in the if else.

I will be need to get approvals or find an alternative way to generate a non standard pdf to add a golden test. I will do that and update the PR.

As for the two two golden reference image GIFs, they were generated when i ran make test so added them to the diff. I have now removed the files.

karthik-bashetty:
Hi @tonimelisma,

Added the tests as requested. I could not add the files i had already as they had confidential data and redacting ended up removing the extra bytes at the start of the PDF. I found the test file attached at https://github.com/pdf-association/pdf20examples/blob/master/PDF%202.0%20with%20offset%20start.pdf.

PR #456, opened 2024-11-20
write icc files to tmp only when necessary (with tests)

### LINKED ISSUE #435: Temporary Directory remains on disk
### Issue Description
When you import govips, a temporary folder is always created, but not always removed.

### Replication Steps:
1.  Import govips library
2. Execute a simple command, such as print(vips.Version) (without invoking vips.Startup()).
3. Allow the program to terminate.

In this scenario, the temporary directory persists on disk, which it should not. The only way to ensure its removal currently is to explicitly call vips.Startup() and then defer vips.Shutdown()

### Real-world Scenario:
- Package A imports govips because it uses some govips functions
- Package **main** imports **Package A**
- Pacakge **main** only uses some constants from **Package A** (not related to govips)
- After performing its tasks, the **main** package completes execution, but the temporary govips folder remains on disk

### Proposed Solutions:
After reviewing the code, I suggest two potential solutions to address this issue:
1. Create the Temporary Directory During vips.Startup(): This would ensure that the temporary folder is only created when vips.Startup() is explicitly called. It is important to note that in the current implementation, the folder is deleted only if the ### **running** variable is set to TRUE (as managed in vips.Shutdown()).
2. Expose a Method to Delete the Temporary Folder: Provide a method within the govips library that can be called to remove the temporary directory explicitly. This would allow users to clean up the directory without relying on vips.Startup() and vips.Shutdown() if they are not using the full capabilities of the library.

rhafer:
This is the same as https://github.com/davidbyttow/govips/pull/442 it just adapts the unit test additionally.

Should fix #435

### DIFF
--- a/vips/govips.go
+++ b/vips/govips.go
@@ -183,7 +183,9 @@ func Shutdown() {
 		return
 	}
 
-	os.RemoveAll(temporaryDirectory)
+	if temporaryDirectory != "" {
+		os.RemoveAll(temporaryDirectory)
+	}
 
 	C.vips_shutdown()
 	disableLogging()
--- a/vips/govips.go
+++ b/vips/govips.go
@@ -96,8 +96,6 @@ func Startup(config *Config) {
 		panic(fmt.Sprintf("Failed to start vips code=%v", err))
 	}
 
-	initializeICCProfiles()
-
 	running = true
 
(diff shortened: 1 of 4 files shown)


Auxority:
Nice work!

PR #464, opened 2025-02-24
Add some functions

Flash-nDie:
Hello. 

### PR Description 
This PR adds several functions corresponding to the libvips C API.

### Added Functions
- func (r *ImageRef) Background() ([]float64, error) : Retrieves the background colors.
- func (r *ImageRef) Subtract(in2 *ImageRef) error : Performs image subtraction.
- func (r *ImageRef) Abs() error : Applies the absolute value operation.
- func (r *ImageRef) Project() (*ImageRef, *ImageRef, error): Executes the project operation.

### Modified Function
The text() function has been revised:
- The original implementation is no longer used.
- Since text() belongs to the create operations, it has been moved to create.c.
- func Text(params *TextParams) (*ImageRef, error) : Create text image.

### DIFF
--- a/vips/arithmetic.c
+++ b/vips/arithmetic.c
@@ -81,3 +81,7 @@ int absOp(VipsImage *img, VipsImage **out) {
 int project(VipsImage *in, VipsImage **col, VipsImage **row) {
   return vips_project(in, col, row, NULL);
 }
+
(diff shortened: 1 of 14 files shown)


tonimelisma:
Thank you so much @Flash-nDie!

Issue #465, opened 2025-03-11
Implicit BMP to PNG conversion using golang.org/x/image/bmp decoder causes decode failures for Govips library


jhford-scout24:
I have a patch incoming, just need to sort out some issues with the test suite which occur whether my patch is applied or not.

Thank you for all the great work on the v2 release of govips.  It's now running in production for us and has resolved a huge number of issues we were experience, including segfaults and memory leaks.  It's also increased performance substantially.

There's a problem that govips is currently using the `golang.org/x/image/bmp` to pre-process all BMP files.  The issue with this is that BMP files written to the OS/2 1.x format or any version of BMP which uses RLE compression cannot be decoded by the golang bmp implementation.  Vips under the hood is completely able to handle all of these old BMP format images including RLE encoded using the magick loader.  

In the case that someone is trying to handle a BMP image without linking to magick, it's really simple for them to use the `golang.org/x/image/bmp` implementation in a couple lines.  In the inverse case, where I specifically don't want to use this bmp loader, I need to fork this library and maintain it myself.

The relevant test BMP images can be generated using ImageMagick:
```bash
#!/bin/bash

magick koala.png -type Palette -compress RLE koala-rle.bmp
magick koala.png -type Palette -compress RLE koala-rle.bmp2
magick koala.png -type Palette -compress RLE koala-rle.bmp3

magick koala.png koala.bmp
magick koala.png koala.bmp2
magick koala.png koala.bmp3
```
And we can see that they are processed by a full-featured version of Vips: 

```bash
#!/bin/bash

for i in koala{,-rle}.bmp{,3,2} ; do
  ${VIPS:-vips} copy $i $i-output.png
  file $i-output.png
done
```
```
~/scout24/govips $ ./test-bmps.sh
koala.bmp-output.png: PNG image data, 1024 x 768, 8-bit/color RGB, non-interlaced
koala.bmp3-output.png: PNG image data, 1024 x 768, 8-bit/color RGB, non-interlaced
koala.bmp2-output.png: PNG image data, 1024 x 768, 8-bit/color RGB, non-interlaced
koala-rle.bmp-output.png: PNG image data, 1024 x 768, 8-bit/color RGB, non-interlaced
koala-rle.bmp3-output.png: PNG image data, 1024 x 768, 8-bit/color RGB, non-interlaced
koala-rle.bmp2-output.png: PNG image data, 1024 x 768, 8-bit/color RGB, non-interlaced
```

[...]

PR #466, opened 2025-03-11
Stop implicitly converting BMP files to PNG using golang.org/x/image/bmp

jhford-scout24:
This BMP library lacks support for BMPv2 (OS/2 1.x format) and all RLE encoded BMP files.

This patch defers all BMP handling to the caller.  Detecting and converting a BMP file is quite a simple task which does not introduce any CGo dependencies.

The tests are mostly passing locally for me, however, I am getting segfaults in the heif encoder as well as some errors in corrupted JPEG loading, neither were touched by this patch, so it's unclear what's going on here.


jhford-scout24:
It looks like builds and tests are broken for quite some time.

I believe that the Ubuntu version of libvips is built with magick support, which should be able to load BMP files fine.  I'm not sure how to go about showing that this works enough to merge given that tests are broken.

jhford-scout24:
resolves https://github.com/davidbyttow/govips/issues/465

PR #471, opened 2025-03-19
add keep option when export image

Flash-nDie:
### PR Description 
This PR adds keep option to  Export Params. Since libvips version 8.15, the strip option has been deprecated. Therefore, for versions 8.15 and above, the keep option must be provided. However, for versions below 8.15, the strip option should be maintained for compatibility. To handle this, an #ifdef macro is used to apply the keep option only for libvips 8.15 and above.

### DIFF
--- a/vips/foreign.c
+++ b/vips/foreign.c
@@ -262,9 +262,11 @@ int set_jpegsave_options(VipsOperation *operation, SaveParams *params) {
     ret = vips_object_set(VIPS_OBJECT(operation), "Q", params->quality, NULL);
   }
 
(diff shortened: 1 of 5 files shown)


villahousut:
Hey @Flash-nDie thanks for the PR. It looks like you're overwriting the return value when you set the keep parameter each time. Any errors returned in previous function calls are lost.

I also noticed there's no unit tests nor golden tests. You've only set the parameter for some golden tests.

Flash-nDie:
@tonimelisma 
Hello. I revised the code to check previous error, and added golden test which strip icc profile from image by keep options.

Flash-nDie:
@tonimelisma 
Hello, again. The PR was just closed — is there any way I can know the reason? Or should I open a new one?

PR #452, opened 2024-10-12
add side effects import for image/jpeg png and gif to avoid image: unknown format

BennyThink:
### Summary

The test code below demonstrates an issue where `govips` does not correctly handle JPEG images, resulting in the error: `image: unknown format`.

```go
package main

import (
	"fmt"
	"github.com/davidbyttow/govips/v2/vips"
)

func main() {
	vips.Startup(nil)
	defer vips.Shutdown()
	image1, _ := vips.NewImageFromFile("test.jpg")
	ep := vips.NewDefaultJPEGExportParams()
	_, err := image1.ToImage(ep)
	fmt.Println(err)
}
```

# Issue
The underlying cause is that govips does not automatically import the `image/jpeg` package as a side effect. As a result, JPEG and GIF formats are not recognized, whereas formats like PNG (which are handled [via an import](https://github.com/davidbyttow/govips/blob/master/vips/foreign.go#L9) in `foreign.go` ) work correctly.

# Solution
This PR addresses the issue by explicitly adding the necessary side effect imports for JPEG and GIF formats in `image.go`. With this change, govips will correctly handle JPEG and GIF images without requiring additional user intervention.

PR #483, opened 2025-06-08
add side effects import for golang.org/x/image/webp to avoid image: unknown format on webp image

n0vad3v:
Same issue as https://github.com/davidbyttow/govips/pull/452.


n0vad3v:
Wow, this merge is fast! ❤️

PR #526, opened 2026-04-15
Add NewImageFromHWCArray for load raw pixels

douyun-dtyq:
Allow user to load raw pixels bytes buffer. Convenience for processing data from llm (i.e. onnxruntime) or other apis.

> [!NOTE]
> ### Add `NewImageFromHWCArray` to create a Vips image from raw pixel data
> Adds `NewImageFromHWCArray` in [vips/image.go](https://github.com/davidbyttow/govips/pull/526/files#diff-e488e36196c25a401a5af66dd4e6010e8b4e2dec77d4eaba2b6c66fe86a3933b) to construct an `ImageRef` from a caller-provided HWC-ordered byte slice. The function copies the input buffer, calls the C-level `create_image_from_memory_copy` with width, height, bands, and band format, sets the Vips interpretation, and returns an `ImageRef` with `ImageTypeUnknown` as the format. `runtime.KeepAlive` is used to prevent the copied buffer from being GC'd during the C call.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 11a7afc.</sup>
> 
>

### DIFF
--- a/vips/image.go
+++ b/vips/image.go
@@ -715,6 +715,34 @@ func NewImageFromGoImage(img image.Image) (*ImageRef, error) {
 	return newImageRef(vipsImage, ImageTypeUnknown, ImageTypeUnknown, nil), nil
 }
 
+// NewImageFromHWCArray creates a new ImageRef from a pixel data in HWC(height, width, channels) format.
+func NewImageFromHWCArray(pixels []byte, bands int, width, height int, bandFormat BandFormat, interpretation Interpretation) (*ImageRef, error) {
+	if err := startupIfNeeded(); err != nil {
+		return nil, err
+	}
+
+	// copy pixels
+	_pixels := make([]byte, len(pixels))
+	copy(_pixels, pixels)
+
+	vipsImage := C.create_image_from_memory_copy(
+		unsafe.Pointer(&_pixels[0]),
+		C.size_t(len(_pixels)),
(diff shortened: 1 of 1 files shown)


davidbyttow:
Thanks for the contribution. The underlying idea is useful and the call pattern mirrors `NewImageFromGoImage` correctly, but I don't want to land new public API without a few things addressed first, so I'm closing this for now. Happy to look at a follow-up PR that covers:

1. **Tests.** At minimum a round-trip test that constructs an `ImageRef` from a known HWC byte slice and verifies width/height/bands/interpretation. Public API needs coverage.
2. **Drop the redundant `buf` arg to `newImageRef`.** `create_image_from_memory_copy` already copies the data, so the parallel `NewImageFromGoImage` passes `nil` there. Passing `_pixels` is inconsistent and keeps a Go-side buffer alive for no reason.
3. **Idiomatic naming.** `_pixels` → `pixelsCopy` (or just inline the copy).
4. **Input validation.** Reject empty buffers and ideally check `len(pixels) == width*height*bands*sizeofBandFormat` so callers get a clear error instead of a libvips crash.

Reopen with those and I'll merge.

douyun-dtyq:
I'm too busy lately to follow up on this. Just consider this PR an idea share or reference. Feel free to take over or use the code however you want.

PR #441, opened 2024-06-28
add function LoadImageFromBytes

code1super:
for load again byte array received from image.ToBytes() 
it good for fast process pixels of image

PR #528, opened 2026-06-07
Add NewImageFromMemory for raw pixel buffers

### LINKED ISSUE #480: API to create a vips image from go image ?
Does one exist ? The other way around is simple enough via ToImage(), but how does one wrap a vips image over an existing go one or import it ? Surely there's a faster way than encoding the go image as jpg/png to memory and then load it with vips.

davidbyttow:
## Summary

Exposes `vips_image_new_from_memory_copy` through a Go-friendly API:

```go
ref, err := vips.NewImageFromMemory(pixels, width, height, bands, vips.BandFormatUchar, vips.InterpretationSRGB)
```

Fills the gap between `NewImageFromGoImage` (forces an `image.Image` / NRGBA boundary) and `NewImageFromBuffer` (encoded PNG/JPEG/etc.). Motivating use cases: raw pixel data from ML inference (ONNX, TensorFlow outputs), camera sensors, or any in-memory layout the caller can describe.

Closes #480. Also addresses the broader ask in #441 / #526.

## Behavior

- Up-front validation of dimensions, band format, and buffer length — callers get a clear Go error instead of a libvips crash.
- Pixel data is copied into libvips (`create_image_from_memory_copy`), so the input slice is safe to reuse or free immediately after the call returns.
- Supports all real BandFormats (uchar/char/ushort/short/uint/int/float/double). Complex/DPComplex are explicitly rejected — they're rarely useful here and the sample sizing is ambiguous.

## Test plan

- [x] `TestNewImageFromMemory_RGB` — 2x2 RGB round-trip, verifies width/height/bands/format/interpretation and corner pixel values.
- [x] `TestNewImageFromMemory_Float` — 1x1 single-band float image, verifies float decoding.
- [x] `TestNewImageFromMemory_BufferIsCopied` — mutates the caller-side slice after construction and confirms the vips image is unchanged.
- [x] `TestNewImageFromMemory_InvalidArgs` — zero width / zero height / zero bands / size mismatch / unsupported band format all return errors with no `ImageRef`.
- [x] Existing `TestNewImageFromGoImage*` still pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

> [!NOTE]
> ### Add `NewImageFromMemory` to create a VipsImage from a raw pixel buffer
> - Adds `NewImageFromMemory` in [vips/image.go](https://github.com/davidbyttow/govips/pull/528/files#diff-e488e36196c25a401a5af66dd4e6010e8b4e2dec77d4eaba2b6c66fe86a3933b) that constructs an `ImageRef` from a contiguous, uncompressed pixel buffer given width, height, bands, `BandFormat`, and `VipsInterpretation`.
> - Validates dimensions, checks bytes-per-sample via a new `bandFormatSizes` map (complex formats unsupported), and verifies the buffer length matches the expected size before calling into libvips.
> - The input buffer is copied via `C.create_image_from_memory_copy`, so callers can safely mutate or reuse it after creation.
> - Tests cover RGB and float single-band images, buffer isolation, and all invalid-argument error paths.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6ad97c8.</sup>
> 
>
>

### DIFF
--- a/vips/image.go
+++ b/vips/image.go
@@ -715,6 +715,70 @@ func NewImageFromGoImage(img image.Image) (*ImageRef, error) {
 	return newImageRef(vipsImage, ImageTypeUnknown, ImageTypeUnknown, nil), nil
 }
 
+// NewImageFromMemory creates a new ImageRef from a raw, uncompressed pixel
+// buffer. The buffer is interpreted as height rows of width pixels, each pixel
+// holding bands samples of the given BandFormat, laid out contiguously (no row
+// stride padding).
+//
+// Use this for raw byte arrays from sources like ML inference outputs (ONNX,
+// TensorFlow), camera sensors, or any in-memory pixel data. For Go's
+// image.Image, use NewImageFromGoImage. For compressed bytes (PNG, JPEG, etc.),
+// use NewImageFromBuffer.
+//
+// The pixel data is copied into libvips, so the caller is free to reuse or
+// free the input slice after this call returns.
+func NewImageFromMemory(pixels []byte, width, height, bands int, format BandFormat, interpretation Interpretation) (*ImageRef, error) {
+	if err := startupIfNeeded(); err != nil {
+		return nil, err
+	}
+
+	if width <= 0 || height <= 0 || bands <= 0 {
+		return nil, fmt.Errorf("invalid dimensions: width=%d height=%d bands=%d", width, height, bands)
+	}
+
+	bytesPerSample, ok := bandFormatSizes[format]
+	if !ok {
+		return nil, fmt.Errorf("unsupported band format: %d", format)
+	}
+
+	expected := width * height * bands * bytesPerSample
+	if len(pixels) != expected {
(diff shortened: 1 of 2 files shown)


davidbyttow:
CI is red but the failures are pre-existing infrastructure rot, not caused by this PR:

```
heif: Unsupported feature: Unsupported codec (4.3000)
```

All 4 failing tests (`TestImage_AutoRotate_6__heic_to_jpg`, `TestImage_Add`, `TestImage_Multiply`, `TestImageRef_Divide`) load `heic-24bit.heic`. Ubuntu split the HEVC decoder out of `libheif1` into separate plugin packages (`libheif-plugin-libde265` / `libheif-plugin-x265`) around 24.04, and the workflow only installs `libvips-dev`, which no longer pulls a HEVC decoder transitively. `ubuntu-latest` updated between the last green master run (May 13) and now, surfacing the gap.

Opening a separate one-line workflow fix that adds the plugin to the apt install step. Once that lands, this PR's CI will be green on re-run.

PR #530, opened 2026-06-12
Add streaming I/O: LoadImageFromReader, SaveToWriter, TranscodeStream

antst:
## Summary

Adds streaming I/O via libvips `VipsSourceCustom`/`VipsTargetCustom`, so images can be loaded from any `io.Reader` and saved to any `io.Writer` without holding the full compressed input or output in Go memory — plus an end-to-end streaming transcode pipeline. All new API is purely additive; output bytes are identical to the corresponding `Export*` methods.

```go
// Stream an HTTP upload through a transcode into storage, hashing while writing
err := vips.TranscodeStream(req.Body, io.MultiWriter(hasher, dst), &vips.TranscodeOptions{
    Format:       vips.ImageTypeJPEG,
    AutoRotate:   true,
    ExportParams: &vips.ExportParams{Quality: 85, Interlaced: false},
})
```

## New API

- **`LoadImageFromReader(r io.Reader, params *ImportParams) (*ImageRef, error)`**
  - Default: the image is materialized before returning (memory at or below a disc threshold, an unlinked scratch file above it) and the reader is released immediately — same lifecycle as the buffer path.
  - `params.Access = AccessSequential`: fully lazy streaming. libvips pulls compressed bytes on demand while later operations consume pixels; peak RAM is bounded by line caches, independent of file size and pixel count. The reader stays connected until `Close`.
  - If `r` implements `io.Seeker`, the seek callback is exposed for random-access loading (HEIF etc.); otherwise libvips uses sequential mode with header buffering bounded by **`SetPipeReadLimit`**.
- **`(*ImageRef).SaveToWriter(w io.Writer, format ImageType, params *ExportParams) error`** — JPEG, PNG, WebP, HEIF, TIFF, GIF. Encoded chunks are pushed to `w` as they are produced (TIFF is encoded in memory first: libtiff requires seekable output). `io.Writer` is the chunk callback — composes with `io.MultiWriter`, `io.Pipe`, `io.Copy`.
- **`TranscodeStream(r io.Reader, w io.Writer, opts *TranscodeOptions) error`** — one-shot reader→writer pipeline that picks the decode strategy from header metadata: stays sequential unless `AutoRotate` meets an EXIF orientation ≥3 (90°/180°/270° families), which forces materialization.
- **`SetStreamDiscThreshold(bytes int64)`** / **`SetStreamScratchDir(dir string)`** — knobs for the materialization path (defaults: `VIPS_DISC_THRESHOLD` env or 100 MB; `os.TempDir()`). Scratch files are unlinked as soon as they are reopened, so they cannot leak even on crash.

## Design

- Callback bridge: a global registry of integer handles → reader/writer entries, with static C trampolines dispatching to exported Go callbacks (the standard cgo pattern). Per-instance mutexes serialize callbacks, so callers never need thread-safe readers/writers. Handles wrap within int32 range (they cross CGo as C `int` via `GINT_TO_POINTER`).
- Loader auto-detection via `vips_foreign_find_load_source`; both load and save reuse the existing `set_*load_options` / `set_*save_options` from foreign.c, and the Go side shares the `SaveParams` builders with the buffer savers — so streaming output cannot drift from `Export*`.
- Streaming operations build uncached (`vips_object_build`): their cache key contains the unique source/target object, so hits are impossible and caching would only pin sources and lazy images.
- Truncated/network-failed streams error deterministically (`fail_on=truncated`): at load time on the materialized path, at the first full pixel pass on the sequential path — wrapped with the original reader error (`errors.Is`-matchable).

## Tests & numbers

- Byte identity vs `Export*` for all six formats on every path (buffer-materialized, disc-materialized, sequential).
- Pixel bomb: 1.7 MB compressed → 144 MB decoded streams under a tracked-memory ceiling of decoded/4; sequential load of that input returns after reading **8 KB**.
- Incremental pacing: first output chunk arrives long before the input is consumed; output arrives in ~100 chunks during encode.
- Backpressure: a writer stalling 2 s mid-encode causes bounded memory growth, no deadlock, byte-identical output.
- Concurrency under `-race`, leak checks (registry empty + no open ImageRefs), golden-file comparison.
- Benchmarks (2.9 MB TIFF): `BenchmarkStreamLoad` 1,271 B/op vs `BenchmarkBufferLoad` 2,910,479 B/op of Go allocations.

## Notes for reviewers

- The committed golden file is the macOS/arm64/libvips-8.18 variant; Linux CI will generate its platform-specific golden on first run (same naming scheme as the existing goldens).
- Documented caveat worth knowing: progressive (interlaced) JPEG cannot stream in either direction — and the package's default JPEG export params set `Interlace: true`, so streaming consumers should pass `Interlaced: false`. HEIF decodes whole-frame inside libheif regardless of access mode. The README's new streaming section includes the sequential-safe vs. materializing operation table.
- Also includes a drive-by fix: the `FindTrim` tests leaked their `ImageRef`s (relevant since the leak detector landed in #525).

> [!NOTE]
> ### Add streaming I/O via `LoadImageFromReader`, `SaveToWriter`, and `TranscodeStream`
> - Adds `LoadImageFromReader` to load images from an `io.Reader`, supporting sequential lazy mode or immediate materialization to memory or disc based on decoded size vs. a configurable threshold (`SetStreamDiscThreshold`).
> - Adds `ImageRef.SaveToWriter` to stream encoded output to an `io.Writer` for JPEG, PNG, WebP, HEIF, and GIF; TIFF falls back to a single-chunk buffer write.
> - Adds `TranscodeStream` for an end-to-end decode→transform→encode pipeline between `io.Reader` and `io.Writer`, selecting sequential or materialized path based on `AutoRotate`/EXIF orientation.
> - Refactors per-format buffer savers in [vips/foreign.go](https://github.com/davidbyttow/govips/pull/530/files#diff-ec94a2793e4149e5e3c5e926e52086a17d586f71fe03e67405c96512de0d6181) to use shared `newSaveParams*` helpers, reused by the new streaming savers.
> - Risk: sequential saves surface decode errors late (during the save operation); callers must check returned errors from `SaveToWriter` and `TranscodeStream` even after a successful load.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized dbcc405.</sup>
> 
>


davidbyttow (changes requested):
Thorough deep-dive review (code trace + local test runs on macOS arm64, libvips 8.18.0). The core design is solid and the test suite is unusually strong — byte-identity across formats/paths, the pixel bomb, backpressure, truncation determinism, and leak assertions all pass locally, including under `-race` (only HEIC-input tests fail here, environmental: this machine's vips-heif module can't dlopen). I traced the callback bridge end to end and found no memory-safety bug: handle wraparound/skip logic is correct, stale handles fail safe to a vips error rather than a UAF, refcount ordering is right on every error branch, and the leak-detector interaction is clean.

Requesting changes for one required item plus a few worth fixing while you're in there:

**Required: rebase onto master — `vips/foreign.go` now conflicts semantically with #535.** #535 added a version-guarded `TargetSize` block inside `vipsSaveWebPToBuffer` — exactly the body this PR extracts into `newSaveParamsWebP`. The guard returns an error (`TargetSize requires libvips 8.17.4+`), but `newSaveParamsWebP` returns `(params, cleanup)` with no error. Taking either side loses something: the helper needs an error return (or a separate validation step), and the guard must cover both the buffer and streaming WebP paths.

**API: `SaveToWriter` only accepts the deprecated-style generic `ExportParams`.** That means no streaming access to PNG `Palette/Dither/Bitdepth/Filter`, WebP `NearLossless/IccProfile/TargetSize`, TIFF `Predictor/Tile`, HEIF `Bitdepth/Effort`, or GIF `Dither/Effort/Bitdepth` — even though the C plumbing fully supports them. This is much harder to change after it ships; I'd take format-specific params (matching the `Export*` API) now.

**Docs/semantics:**
- Streaming load rewrites `fail=true` to `fail_on=truncated`, which is less strict than the buffer loaders' `fail_on=error`. Defensible, but it should be in the `LoadImageFromReader` godoc, not just a C comment.
- The "scratch files cannot leak even on crash" claim is overstated: the unlink happens after the full render pass, so a crash mid-materialization leaks the temp file. The happy-path guarantee is real and tested; scope the claim to that.
- `NewImageFromReader` (slurps) vs `LoadImageFromReader` (streams) needs cross-referencing doc comments — that pair is a user trap.

**Smaller:**
- `sourceRead` busy-spins forever on a reader returning `(0, nil)` — inside a cgo callback holding the entry mutex. A bounded retry that fails safe would be better.
- `sourceSeek`'s past-EOF `SEEK_CUR` doesn't match the "exactly POSIX" comment (resolves from EOF, not the reported virtual position). Low practical risk (libheif seeks absolutely), but fix the comment or track a virtual position.
- `save_tiff_to_target` is dead code (Go always takes the TIFF buffer fallback).
- `takeErr()` reads without clearing — rename or actually clear.
- `streamSaveParams` hand-duplicates the `ExportParams`→format-params mapping from the deprecated `Export`; they match today (verified field by field) but it's a drift hazard — extract shared mapping funcs.

Really appreciate the investigation quality here (and in #531, now merged). Happy to re-review quickly after the rebase.

davidbyttow:
Thanks again for this work — rather than wait on a rebase round-trip, I've landed it via #539: your branch merged with authorship preserved, plus the review items (WebP TargetSize guard across both save paths, typed SaveToWriter variants, shared Export mapping helpers, bounded sourceRead retries, doc fixes, dead TIFF target saver removed). Closing this in favor of #539.

PR #537, opened 2026-07-25
Expose image evaluation kill flag

goodmartian:
## Summary

- expose `ImageRef.SetKill(bool)` for cooperative cancellation of libvips evaluation
- serialize the signal with `Close` and treat an already closed image as a no-op
- leave image replacement semantics unchanged
- cover setting, clearing, and closed-image behavior

## Tests

- `go test -race -count=20 ./vips -run '^TestImageRef_SetKill'`
- `go test -count=1 -run '^$' ./...`
- `go vet ./...`

## Local full-suite note

On macOS with Go 1.26.4 and libvips 8.18.4, `go test -race -count=1 ./vips`
hits the same native crash on this branch and on unchanged current `master`
(`eec4bbd`). The targeted race tests, all-package compile gate, and vet gate
above pass.

### DIFF
--- a/vips/image.go
+++ b/vips/image.go
@@ -814,6 +814,16 @@ func (r *ImageRef) Close() {
 	r.lock.Unlock()
 }
 
+// SetKill sets the libvips kill flag used to stop image evaluation.
+func (r *ImageRef) SetKill(kill bool) {
+	r.lock.Lock()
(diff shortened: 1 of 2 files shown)


davidbyttow:
Thank you @goodmartian

PR #543, opened 2026-08-13
Add runtime.KeepAlive for ImageRef arguments in two-image ops

antst:
## Summary

`ImageRef` carries a finalizer that unrefs the underlying `VipsImage`:

```go
runtime.SetFinalizer(imageRef, finalizeImage)   // -> ref.Close() -> g_object_unref
```

The two-image methods already guard the **receiver** with `defer runtime.KeepAlive(r)`, but not the `*ImageRef` they take as an **argument**. Once the callee has loaded `arg.image`, nothing in Go refers to `arg` any more, so the GC is free to finalize it — dropping the `VipsImage` reference while the cgo call is still using that pointer. This is exactly the case [`runtime.KeepAlive`'s documentation](https://pkg.go.dev/runtime#KeepAlive) describes.

It's reachable from ordinary code whenever the argument's last reference is the call expression itself:

```go
base.Composite(loadOverlay(), BlendModeOver, 0, 0)
```

The receiver and the argument are exposed to the same hazard, so this guards both. Twelve sites:

| file | methods |
|---|---|
| `image_pixel.go` | `Mapim`, `Maplut`, `BandJoin`, `Add`, `Multiply`, `Divide`, `Subtract` |
| `image_composite.go` | `CompositeMulti`, `Composite`, `Insert`, `Join`, `ArrayJoin` |

## Honest scope

**I could not reproduce a crash from this.** I ran a targeted loop (argument deliberately unreachable during the call, sibling goroutine spinning `runtime.GC()`, `GOGC=5`, Guard Malloc) for 400 iterations with no fault. The window between the Go-side field load and libvips taking its own ref is a handful of instructions, so that's expected — this is a latent hazard, not a demonstrated failure.

I'm raising it as a consistency fix rather than a bug fix: the receiver-side `KeepAlive` calls already in these functions imply the hazard is understood, and the argument is the same hazard left unguarded. Zero behavior change, and no cost — `KeepAlive` emits no instructions, it only extends the liveness range the compiler must honor.

## Related

`fix/keepalive-cgo` (yours, Feb) does the receiver side in `image.go` and is still unmerged. The two don't overlap — that branch covers `Bands()`, `ColorSpace()`, `ImageFields()` and friends in `image.go`; this covers argument positions in `image_pixel.go` / `image_composite.go`. Happy to fold that branch in here and send a single complete pass instead, if you'd prefer.

I also have an env-gated stress harness (`GOVIPS_STRESS=1`) left over from the #531 investigation — the Guard Malloc / `GOGC` rig that isolated the `Join` double-unref. Glad to add it in a separate PR if it'd be useful; leaving it out here to keep this reviewable.

## Verification

`go build ./...`, `go vet ./vips/`, and the full `go test ./vips/` suite pass on macOS arm64 / libvips 8.18.4. `gofmt` clean (the two pre-existing unformatted files, `generated.go` and `image_test.go`, are untouched).

> [!NOTE]
> ### Add `runtime.KeepAlive` for secondary `ImageRef` arguments in multi-image operations
> Without `runtime.KeepAlive`, the Go GC can finalize `ImageRef` arguments before a CGo call completes, causing use-after-free bugs. Adds `defer runtime.KeepAlive(...)` for every secondary image argument in [image_composite.go](https://github.com/davidbyttow/govips/pull/543/files#diff-d28522bd51ae2b70c3bbbdcfdcba392aa58c7b7f004518f0e707749b33105721) and [image_pixel.go](https://github.com/davidbyttow/govips/pull/543/files#diff-c8d930dc147921b711478df2c595ababbe2668169248e5df14e9caf8f97b9c73), covering `Composite`, `CompositeMulti`, `Insert`, `Join`, `ArrayJoin`, `Mapim`, `Maplut`, `BandJoin`, `Add`, `Multiply`, `Divide`, and `Subtract`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 448398a.</sup>
> 
>

### DIFF
--- a/vips/image_composite.go
+++ b/vips/image_composite.go
@@ -8,6 +8,7 @@ import "runtime"
 // CompositeMulti composites the given overlay image on top of the associated image with provided blending mode.
 func (r *ImageRef) CompositeMulti(ins []*ImageComposite) error {
 	defer runtime.KeepAlive(r)
+	defer runtime.KeepAlive(ins)
 	out, err := vipsComposite(toVipsCompositeStructs(r, ins))
 	if err != nil {
 		return err
@@ -19,6 +20,7 @@ func (r *ImageRef) CompositeMulti(ins []*ImageComposite) error {
 // Composite composites the given overlay image on top of the associated image with provided blending mode.
 func (r *ImageRef) Composite(overlay *ImageRef, mode BlendMode, x, y int) error {
 	defer runtime.KeepAlive(r)
+	defer runtime.KeepAlive(overlay)
 	out, err := vipsGenComposite2(r.image, overlay.image, mode, &Composite2Options{X: &x, Y: &y})
 	if err != nil {
 		return err
@@ -30,6 +32,7 @@ func (r *ImageRef) Composite(overlay *ImageRef, mode BlendMode, x, y int) error 
 // Insert draws the image on top of the associated image at the given coordinates.
 func (r *ImageRef) Insert(sub *ImageRef, x, y int, expand bool, background *ColorRGBA) error {
 	defer runtime.KeepAlive(r)
+	defer runtime.KeepAlive(sub)
 	insertOpts := &InsertOptions{Expand: &expand}
 	if background != nil {
 		insertOpts.Background = []float64{float64(background.R), float64(background.G), float64(background.B), float64(background.A)}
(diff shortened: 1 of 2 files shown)


davidbyttow:
Merged and released in [v2.19.0](https://github.com/davidbyttow/govips/releases/tag/v2.19.0), thanks. One note on the description: `fix/keepalive-cgo` isn't outstanding. It merged back in February as #494 (squashed, which is why the branch still looked unmerged), and those calls moved into the `image_*.go` files with the `ImageRef` split in #525. So there's nothing left to fold in, and I've deleted the stale branch. I'll pass on the stress harness for now, but if another wandering use-after-free turns up I may take you up on it.

